PR #5, opened 2017-01-09
spelling correction for Mozilla in DESCRIPTION

dpastoor:


dpastoor:
thanks - package looks AMAZING!

jjallaire:
Glad you like it! Let me know if there is anything that can be changed or added to make it better.

dpastoor:
Will run it through it's paces before Rstudio Conf so can hopefully give some feedback then. This will be amazing for the graduate school classes and for pre-training before workshops.

PR #9, opened 2017-01-26
enable autocompletion in tutor Ace instances

kevinushey:
NOTE: not yet ready for merge but opening a PR for discussion.

This PR will be used to add autocompletion support to the editors used in an R Markdown tutorial document. Right now we just have the main skeleton necessary for registering a custom completer on Ace; hooking that up to R with a basic completion system is still a WIP.

Now that we're planning different tutor modules (e.g. autocompletion, diagnostics), should we consider generating a single minified `tutor.min.js` from the core `tutor.js` stuff alongside the `tutor-autocompletion.js` file + (pending) diagnostics?


kevinushey:
I think this PR should be good to go -- @jjallaire, can you review + merge if all looks well? We can iterate on specific behaviors on master if there's anything you feel can / should be tweaked.

inst/lib/tutor/tutor-autocompletion.js
```diff
@@ -0,0 +1,209 @@
...
+
+    // manually handle event
+    event.stopPropagation();
+    event.preventDefault();
+    editor.execCommand("startAutocomplete");
+  };
+
+  document.addEventListener("keydown", function(event) {
+
+    // TODO: find more appropriate place for one-time initialization
```
jjallaire: Are we okay w/ this or is there still an issue outstanding here?

sandbox/autocompletion.html
```diff
@@ -0,0 +1,291 @@
+<!DOCTYPE html>
```
jjallaire: We should just .gitignore this file

jjallaire:
I got this error in the R console when typing within the autocompletion sandbox Rmd:

```
Listening on http://127.0.0.1:5678
Warning: Error in eval: could not find function "<-"
Stack trace (innermost first):
    3: <Anonymous>
    2: do.call
    1: rmarkdown::run [/Users/jjallaire/packages/rmarkdown/R/shiny.R#203]
Error : could not find function "<-"
```

jjallaire:
I'm not seeing completions for the setup chunk (i.e. if I type `autocomp|` I don't see completions).

jjallaire:
Looks great! I'll merge it now. One thing I noticed is that we don't get automatically triggered completions for `::` and `$` (the latter being much more important of course). Would that be straightforward to do?

kevinushey:
It should be relatively straightforward -- I'll give that a shot!

jjallaire:
I'm also wondering if we should leave off explicit S3 methods for e.g. `summary` (because that ends up dominating the listing).

kevinushey:
We could, but note that those methods are indeed exported + visible on the search path (ie, they're used both in S3 dispatch and are exported as though they are free functions). We also show those completions in the RStudio IDE and nobody has complained thus far.

jjallaire:
Okay let's leave as-is then!

jjallaire:
A couple more small things to consider:

1) We currently show the "R" scope on the far right but as far as I can see
it's always "R" and never anything else. Seems like we could drop this.

2) The overall width could I think be trimmed to accommodate the contents
better (sort of like we do in RStudio).

Hopefully both of the above could be accomplished w/ just CSS changes

PR #10, opened 2017-01-28
Provide chunk option and documentation for exercise autocompletion

jjallaire:
@kevinushey 

@garrettgman and @dtkaplan we've added editor autocompletion for exercise chunks. When discussing this originally with Garrett he asked that it would be nice to be able to disable autocompletion for certain types of exercises / tutorials. You can do this via the `exercise.completion` option.


kevinushey:
LGTM!

PR #11, opened 2017-01-28
use $ prefix for tutor completer since it's private

jjallaire:
@kevinushey It seems like this variable isn't referenced anywhere so this is the only change necessary, let me know if there is anything I missed.

I prefixed with `$` because `Tutor` actually exposes a public API and I wanted that bit to be private.


kevinushey:
LGTM!

PR #13, opened 2017-01-30
report syntax errors in tutor R chunks

kevinushey:
NOTE: not quite ready for merge; ready for initial review + discussion.

This PR implements a simple diagnostics engine for R code chunks, reporting syntax errors.

screen shot 2017-01-30 at 2 29 14 pm

Currently, only diagnostics related to mismatched bracket pairs, as well as unexpected symbols, are reported -- we should consider what other syntax errors are worth reporting.

This PR is currently not implemented as a web worker, as that unfortunately requires a lot more infrastructure work (we'd need to opt-in to the requirejs-style framework used by Ace), and since diagnostics here are generally only going to be run on small editor chunks, we can reliably generate diagnostics in less than ~200ms -- ergo, we just emit diagnostics synchronously for now.


jjallaire:
I think we should add comma and parentheses oriented diagnostics if those aren't too difficult.

It seems like if we are running this entirely in JS and then only on the currently active editor that diagnostics will run very, very quickly. I think this means that we can run them pretty aggressively (perhaps always after 200ms of inactivity or something like that).

We should also implement an option to disable diagnostics (implementation should mirror disabling of completion).

@dtkaplan has been using sequences of underscores (e.g. `plot(_____)` as a "fill in" prompt for some exercises. We may want to exempt those errors from processing (as a separate matter we might want to consider converting code with `______` into a snippet which would be a nicer editing experience.

When we reload tutorials we restore previously entered code. This happens here: https://github.com/rstudio/tutor/blob/master/inst/lib/tutor/tutor.js#L1383-L1385. We'll want to run diagnostics then as well.

kevinushey:
Sounds good! Fortunately, `plot(_____)` is already ignored as far as diagnostics go. I'll see if we can add some extra diagnostic output.

jjallaire:
I noticed when playing around with this branch that we don't have automatic parentheses matching. As a result any time you call a function you get the red error indicator until you've filled in the arguments and added the end parentheses. 

I think we should add the brace/parentheses/bracket matching. How hard would that be?

kevinushey:
I've updated Ace to v1.2.6; this version includes the expected behaviours necessary for parenthesis auto-insertion by default.

Issue #44, opened 2017-03-17
Skip section instead of (or in addition to) skip exercise


robbyshaver:
The current tutor api includes a method: skipExercise() and fires an event: progressEvent.event === "exercise_skipped".  However, the UI actually allows the user to skip sections, not exercises.

The code in tutorial-format.js works around this, but it's kind of a hack. It passes in the id of a section to skipExercise, and in the handler for the "exercise_skipped" event, fishes the id from the empty jquery object like so:

  function exerciseSkipped(exerciseElement) {
    var sectionSkippedLabel;
    if (exerciseElement.length) {
      console.log('something changed in tutor ...');
      return;
    //  sectionSkippedLabel = exerciseElement.attr('data-label');
    }
    else {  // fish out the id of the section that got skipped
      sectionSkippedLabel = $(exerciseElement).selector.split('"')[1];
    }

It would be much cleaner if we had a skipSection method and a "section_skipped" event to work with - particularly if somebody looks at tutorial-format.js and tries to figure out what's going on.

jjallaire:
So it seems like if I just remove `exercise_skipped` entirely and replaced it with `section_skipped` that would both better conform to what the UI actually allows and enable you to clean up the dirty hack. Is that correct?

robbyshaver:
Yep, that's right.

PR #49, opened 2017-03-17
rename exercise_skipped to section_skipped

jjallaire:
Addresses #44 

@robbyshaver This PR renames the exercise_skipped event to section_skipped. As part of the change your `exerciseSkipped` helper function is no longer called (both `section_completed` and `section_skipped` both call `sectionCompleted`): 

https://github.com/rstudio/tutor/blob/feature/section-skipped/inst/rmarkdown/templates/tutorial/resources/tutorial-format.js#L446-L453

Not sure if this is the right change and/or if other refinements are required. Could you try out this PR locally and let me know if everything is working as you expect and whether we need additional refinement?

### DIFF
--- a/R/events.R
+++ b/R/events.R
@@ -48,20 +48,23 @@ question_submission_event <- function(session,
 }
 
 
-exercise_skipped_event <- function(session, label) {
+section_skipped_event <- function(session, sectionId) {
+  
+  # event data
+  event_data <- list(sectionId = sectionId)
   
   # notify server-side listeners
   record_event(session = session,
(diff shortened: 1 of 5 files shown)


robbyshaver:
@jjallaire We care in the UI about the difference between an exercise getting completed vs. getting skipped, so I renamed/updated the exerciseSkipped function to sectionSkipped and call it when we get the section_skipped event. Everything looks to be working right now.

jjallaire:
Great, I just merged the changes.

PR #59, opened 2017-03-23
save and restore current topic

jjallaire:
Intended to address https://github.com/rstudio/tutor/issues/56

@robbyshaver I've added the code required for tutorials to save and restore their hash and scroll position in this PR. This appears to work fine, but I thought you might want to give it a test drive to confirm before we merge it. Note that we are now restoring scroll position which previously I thought too janky but if we are restoring the topic it only seems right to restore the scroll position.

### DIFF
--- a/inst/lib/tutor/tutor.js
+++ b/inst/lib/tutor/tutor.js
@@ -1501,8 +1501,9 @@ Tutor.prototype.$initializeClientState = function(client_state) {
   window.addEventListener("popstate", maybePersistClientState);
   
   // restore hash if there wasn't a hash already
-  if (!window.location.hash && client_state.hash)
-    history.pushState(null, null, client_state.hash);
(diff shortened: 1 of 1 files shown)


robbyshaver:
The save/restore topic is working great. The scroll restore doesn't work because of what I'm doing in tutorial-format.js. When we switch topics, I set the scroll top to 0. I put that in because our #hash navigation was causing the page to scroll up a bit whenever you switched topics and it looked bad. There may be another way around that problem, but I'm not even sure restoring the scroll position is a good thing - I agree that it's kind of wanky. I'm ok if we always take you to the top of the topic  - and if it's not what you want, you likely just want to jump to the end, which isn't much effort.

jjallaire:
Okay, I just commented out the scroll position restoration so I think this is good to go.

PR #71, opened 2017-05-01
enable R code completion and diagnostics by default

jjallaire:
@kevinushey According to our docs these are enabled by default but I don't think we had this working prior to this change. Let me know if the change looks okay to you.


kevinushey:
I believe you're right -- thanks for catching this!

PR #78, opened 2017-05-24
Update path information for package bundling

alex23lemm:
Tiny modification of the description on how how to bundle tutorials using packages. The original description included the `inst` sub-directory. However, `learnr::run_tutorial("mytutorial", package = "mypackage")` won't work this way since `inst` is not part of the `system.file` call :

    > run_tutorial
    function (name, package, shiny_args = NULL) 
    {
        tutorial_path <- system.file("tutorials", name, package = package)

In order to work properly, you need to store your tutorial directories directly underneath the `tutorials` directory.

### DIFF
--- a/docs/publishing.Rmd
+++ b/docs/publishing.Rmd
@@ -18,7 +18,7 @@ learnr::run_tutorial("slidy", package = "learnr")
 
 To bundle a tutorial into an R package and make it available for running via the `run_tutorial` function you should:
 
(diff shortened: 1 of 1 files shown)


jjallaire:
They do indeed need to be in the `inst` directory (anything in the `inst` directory is automatically moved into the package root directory when the package is built by R CMD build). This code works fine:

```r
learnr::run_tutorial("hello", package = "learnr")
```

And the "hello" tutorial is in fact within the inst directory:

https://github.com/rstudio/learnr/tree/master/inst/tutorials

jjallaire:
Yes, this is a side effect of the devtools override of system.file

Issue #84, opened 2017-06-01
Remove duplicate display of second level header?


garrettgman:
Each second level header appears once in the sidebar and once in the content, where it is often followed immediately by a third level header (if you do not follow it with a third level header, the tutorial will reveal the first two sub-sections at once when progressive reveal  is turned on).

I think this scheme is sub-optimal because I sometimes need to come up with a meaningless third level header to make the progressive reveal work right. The three levels of text that result also look a bit disorienting.

Is it possible to remove the duplicate header (see below)? If an author wants the section title to appear at the top of the section, he or she could repeat it in a third level header.

![untitled 001](https://cloud.githubusercontent.com/assets/85014/26695067/767bb1be-46d7-11e7-882c-9179f2f70bbb.png)

````
---
title: "Exploratory Data Analysis"
output: 
  learnr::tutorial:
    progressive: true
    allow_skip: true
runtime: shiny_prerendered
---

```{r setup, include=FALSE}
library(learnr)
```

## Welcome to EDA

## What is EDA?

### Well what is it?

EDA is Exploratory Data Analysis, everyone's favorite tool. 

### The history of EDA

EDA was popularized by John Tukey.

## The EDA mindset

````

jjallaire:
@robbyshaver What do you think?

robbyshaver:
Sounds reasonable - but would it still be what we want if we fixed what sounds like a bug? "(if you do not follow it with a third level header, the tutorial will reveal the first two sub-sections at once when progressive reveal is turned on)"

garrettgman:
That's a good question. I think changing the behavior would alleviate the issue, but I don't want to imply that the behavior is a bug, because I suspect that the behavior may be sensible given how you implemented section tracking. I think either of these would be a good solutionr:

1. Removing the duplicate label, or 
2. Changing the existing behavior from "Progressive reveal will always begin **after** the first complete sub-section demarcated by a third level header" to "Progressive reveal will always begin **before** the first complete sub-section demarcated by a third level header"

jjallaire:
Garrett if we fixed that bug would you be okay with the behavior?

I personally don't mind the redundancy of the header at the top, as I think
the navigation bar might be too subtle as a page title.

garrettgman:
Yes, I'd be completely happy with that.

jjallaire:
[...]

PR #87, opened 2017-06-05
Add a class to hide a section header (and an example of its use).

robbyshaver:
For issue #84.


jjallaire:
We actually already have this via the bootstrap "hidden" class, e.g.

```
 ### Exercise {.hidden}
```

Let's discuss other possibilities realtime when we can find some time this week.

PR #103, opened 2017-07-18
Fixed checker snippet if clause

gsimchoni:
The if clause was missing the `user_code == check_code` expression, this would make the snippet more clear as to what the `checker` function should perform.

### DIFF
--- a/docs/snippets/codefeedback.md
+++ b/docs/snippets/codefeedback.md
@@ -6,11 +6,11 @@ checker <- function(label, user_code, check_code, envir_result, evaluate_result,
     if (is_bad_code(user_code, check_code))
       return(list(message = "Code wasn't right!", correct = FALSE))
     else
-      return TRUE
+      return(TRUE)
   } 
   
-  # this is a fullly evaluated chunk check
-  else if (user_code == check_code) {
+  # this is a fully evaluated chunk check
+  else {
      list(message = "Great job!", correct = TRUE, location = "append")
   }
 }
--- a/docs/snippets/codefeedback.md
+++ b/docs/snippets/codefeedback.md
@@ -10,7 +10,7 @@ checker <- function(label, user_code, check_code, envir_result, evaluate_result,
   } 
   
   # this is a fullly evaluated chunk check
-  else if {
+  else if (user_code == check_code) {
      list(message = "Great job!", correct = TRUE, location = "append")
   }
 }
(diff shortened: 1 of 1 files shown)


jjallaire:
I think you should only need `is_bad_code()` at the pre-check (because once it comes back in the normal check you will have already validated it). Even if you did need it in both places I also think you could have a single instance of the check (since you do it whether `envir_result` is `NULL` or not.

jjallaire:
Looking at the original code, I actually don't see the problem here at all (I might be dense and missing something though!). You don't need `user_code == check_code` at all, in fact that would be a bug as the `check_code` is actually code that you pass on to a checking engine (e.g. it contains various assertions about what the result should be) rather than code that is compared to the `user_code`. So the original example is basically *only* checking code not checking results.

gsimchoni:
Notice I neglected the `user_code == check_code` part myself and retreated to the generic `is_bad_code` function.

Now, the probability of me being dense is much larger, however the following test fails using the original version (minus some typos), and succeeds using the new version. Please tell me what I'm missing, maybe the test isn't valid at all:

The `checker` function is at its old version (+ I added a basic version of `is_bad_code`):
```
is_bad_code <- function(user_code, check_code) {
  gsub("\\s+", "", user_code) != gsub("\\s+", "", check_code)
}

checker <- function(label, user_code, check_code, envir_result, evaluate_result, ...) {
  if (is.null(envir_result)) {
    if (is_bad_code(user_code, check_code))
      return(list(message = "Code wasn't right!", correct = FALSE, location = "append"))
    else
      return(TRUE)
  } else {
    return(list(message = "Great job!", correct = TRUE, location = "append"))
  }
}
```
And suppose we have a simple markdown asking for the answer to `1+1`, and having a `-check` chunk and a `-check-code` chunk:

    How much is 1+1?
    ``` {r filter2, exercise = TRUE, exercise.eval = FALSE, exercise.lines = 1}
    ```
    ```{r filter2-solution}
    2
    ```

    ```{r filter2-check}
    2
    ```

    ```{r filter2-code-check}
    2
    ```

If you input `1` and submit answer, you'll get `Code wasn't right!` and if you input `2` and submit answer you'll get `Great Job!`

HOWEVER: if you delete the last chunk, the `-code-check` chunk, and input `1` and submit answer - you'll still get `Great Job!`! Because `envir_result` is not `NULL` and you're not checking the `user_code` against anything.

Whereas in the new version I'm proposing, you "visit" the `is_bad_code` result again even in this scenario, and you'll get `Code wasn't right!` when submitting `1`.

What am I missing? Is this test not valid? Thanks.

jjallaire:
I think you should be using a code checking package like checkr here (https://github.com/dtkaplan/checkr)

PR #130, opened 2017-11-16
Use the html4 output format instead of the default html5 in Pandoc 2.0

bborgesr:
This is basically a rip off of rstudio/flexdashboard#150


bborgesr:
cc @jcheng5

PR #144, opened 2018-03-03
Fix minor typo

krlmlr:
regsion -> region, I haven't updated the pkgdown.


bborgesr:
Thanks, I'll regenerate before I submit to CRAN

PR #146, opened 2018-03-08
Display full HTML messages if provided by the `incorrect` or the `correct` args to `question() `

### LINKED ISSUE #136: Adding hyperlinks to incorrect message
It's possible to add a hyperlink to the message displayed when an incorrect choice is selected for a multiple choice question, e.g.

```r
question("What number is the letter A in the alphabet?",
         answer("8", message = "See [here](https://en.wikipedia.org/wiki/English_alphabet) and try again."),
         answer("14"),
         answer("1", correct = TRUE),
         answer("23"), 
         allow_retry = TRUE
         )
```

However in most cases I'd like to display the same incorrect message for all wrong answers, so would like to be able to do something like the following.

```r
question("What number is the letter A in the alphabet?",
         answer("8"),
         answer("14"),
         answer("1", correct = TRUE),
         answer("23"), 
         incorrect = "See [here](https://en.wikipedia.org/wiki/English_alphabet) and try again.",
         allow_retry = TRUE
         )
```

This doesn't give an error however the hyperlink does not work. I think it would be a nice feature to add. The current alternative is the following, which requires repeating the message for each option.

```r
question("What number is the letter A in the alphabet?",
         answer("8", message = "See [here](https://en.wikipedia.org/wiki/English_alphabet) and try again."),
         answer("14", message = "See [here](https://en.wikipedia.org/wiki/English_alphabet) and try again."),
         answer("1", correct = TRUE),
         answer("23", message = "See [here](https://en.wikipedia.org/wiki/English_alphabet) and try again."), 
         allow_retry = TRUE
         )
```

bborgesr:
This was a quick fix on the JS side: essentially grab the .html() provided by the `incorrect` (and the`correct`) arguments to `question()`. Previously, we were grabbing the .text(), which stripped the html rich formatting when provided.

@jjallaire, this seemed like low hanging fruit for this release. Let me know if you think that it risks regression or anything else I might have overlooked.

Closes #136.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,9 @@
 learnr 0.9.2
 ===========
 
-* Improved documentation for deploying `learnr` tutorials in Shiny Server. ([142](https://github.com/rstudio/learnr/issues/142))
+* Fixed [#136](https://github.com/rstudio/learnr/issues/136) by displaying full HTML messages (rather than just the text) if provided by the `incorrect` or the `correct` args to `question()`. ([#146](https://github.com/rstudio/learnr/pull/146))
+
+* Improved documentation for deploying `learnr` tutorials in Shiny Server. ([#142](https://github.com/rstudio/learnr/issues/142))
 
 * Fixed a highlight.js issue from rmarkdown 1.8. ([#133](https://github.com/rstudio/learnr/issues/133))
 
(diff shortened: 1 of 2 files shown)


bborgesr:
@mine-cetinkaya-rundel, can you confirm that installing this branch fixes your use case from #136: `remotes::install_github("rstudio/learnr@fix136")`?

mine-cetinkaya-rundel:
yup, it's working for me too, thanks!

PR #153, opened 2018-03-25
Fix 500 errors when reloading tutorial

bborgesr:
We do this by allowing `$serverRequest()` calls only after the Shiny server has been fully initialized; our method to guarantee this for `learnr` is to send a custom message at the end of the `register_http_handlers()` function (which is in R), that works as a "bat signal" for the JS side to know when to proceed with calls to `$serverRequest()`.

### DIFF
--- a/R/http-handlers.R
+++ b/R/http-handlers.R
@@ -1,7 +1,6 @@
 
 
 register_http_handlers <- function(session, metadata) {
-  message("Got to session", session$token)
   # parent environment for completions (see discussion in setup_exercise_handler
   # for why this is chosen as the completion/execution parent)
   server_envir <- parent.env(parent.env(parent.frame()))
@@ -221,6 +220,7 @@ register_http_handlers <- function(session, metadata) {
     
   }))
   
+  session$sendCustomMessage("isTutorialReady", "true")
 }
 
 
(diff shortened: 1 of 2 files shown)


jjallaire:
This looks good. The only curiosity is that we chose to use a global variable rather than a private within the `Tutorial` class but I'm probably missing some aspect of the orderings/dependencies that requires we go outside the class.

jjallaire:
This tutorial still has a problem with the Continue button: https://jjallaire.shinyapps.io/learnr-tutorial-03a-data-manip-filter/

I actually can't get any of the server-side interations to work in that tutorial

jjallaire:
Wierdly another tutorial I deployed which I thought had fixed the continue button is now broken. I did an update to the pkgdown dev version which also "upgraded" me to the dev versions of like 20 other packages so who knows what the cause is. Let me try deploying learnr stuff from a clean start.

jjallaire:
Okay, I have got all the packages installing from CRAN (I've added packrat to the examples dir so you can reproduce) and none of the first 3 example tutorials (00, 01, and 03a) work at all when you revisit them on shinyapps.io.

jcheng5:
For the record, ^ was due to a corrupted IndexedDB (so corrupted that Chrome could neither display nor delete it in devtools)

Issue #162, opened 2018-05-30
Inconsistent scoping rules for exercise chunks


garrettgman:
I've noticed that the code run in learnr exercise chunks has access to different environments depending on how the tutorial is launched. I expected the code in the exercise chunks to only have access to objects that are created in the setup chunks that are associated with the exercises (as well as the global setup chunk). I see different behavior in two cases:

1. Exercise chunks inherit the global environment of the local R session from which the tutorial is launched when the tutorial is launched from the command line with `learnr::run_tutorial()`. By this I mean that code run in the exercise chunks can use objects stored in the local global environment.

2. Exercise chunks inherit the knit environment of the document that they appear in when tutorial is launched with the GUI Run Document button. By this I mean that code run in an exercise chunk can use objects created in any of the document's normal knitr chunks. This may have something to do with pre-rendering because I do not notice the behavior the first time that I run the document, but on every subsequent run. So far, I've only seen this in local mode (launched from GUI), not on tutorials published on shinyapps.io.

I've hosted a learnr tutorial that describes and demo's these bugs in an attempt to provide a reproducible example: https://tutorials.shinyapps.io/scoping-rules/

jcheng5:
Maybe related? https://github.com/rstudio/learnr/issues/170

schloerke:
So for exorcises in the a learnr document... 
* Setup
* A
* B
* C depends on A

It should make an environment tree like:
```
* Global Environment
* Blank Environment
   `-- Setup
      |-- A
      |   `-- C
      `-- B
```

Currently, it is executing exercise C user interactive code chunks with:
```
* Global Environment
   `-- Setup
      |-- A
      |   `-- C
      `-- B
```
... which should be pulling from a blank environment and not the global environment. (Addressing bullet # 2)


---------------------------------------

Independently, there is a separate issue in that when rendering the document, it should ALWAYS be rendered in a clean environment (like the RStudio IDE does) regardless of where it is rendered.  (Addressing bullet # 1)

-----------------------------

Correct?

schloerke:
To address bullet # 1 : https://github.com/rstudio/rmarkdown/issues/1204 

Gabor addresses the situation by having callr invoke rmarkdown render: https://github.com/rstudio/rmarkdown/issues/1204#issuecomment-344884823
This code can be adopted within learnr with the added dependency of callr

schloerke:
[...]

PR #174, opened 2018-07-27
safe expression evaluation with callr

schloerke:
Address scoping issues within #162 

Pre rendered documents can now be rendered in a clean environment using:
```r
learnr::safe(rmarkdown::render(FILE))
```

Tutorials should be able to be run in a clean environment using:
```r
learnr::safe(learnr::run_tutorial(TUTORIAL))
```

This is done as rmarkdown has no intent on adding a new session for compilation. rstudio/rmarkdown#1204


schloerke:
- [ ] NEWS item
- [ ] test

R/run.R
```diff
@@ -1,39 +1,98 @@
...
     withr::with_envvar(c(RMARKDOWN_RUN_PRERENDER = "0"), {
-      rmarkdown::run(file = NULL, dir = tutorial_path, shiny_args = shiny_args)
+      render_fn <- if (isTRUE(safe_session)) run_safe else rmarkdown::run
+      render_fn(file = NULL, dir = tutorial_path, shiny_args = shiny_args)
     })
   })
 }
+
+
+safe_env <- function() {
```
wch: It would be helpful to have a comment of why this function exists (because it's the same as the default for `callr::r` except that it allows opening a browser).

schloerke:
`callr` does not want to include the `safe` function.  Will keep within `learnr`

R/run.R
```diff
@@ -1,39 +1,109 @@
...
+#' Using \code{safe} should only be necessary when locally deployed.
+#'
+#' @param expr expression that contains all the necessary library calls to
+#'   execute.  Expressions within callr do not inherit the existing,
+#'   loaded libraries.
+#' @export
+#' @examples
+#' \dontrun{
+#' # Direct usage
+#' safe({learnr::run_tutorial("slidy", package = "learnr")})
```
wch: I think having `{ }` around code that's used as an expression is good for some contexts (e.g., shiny and profvis), but in this case, it might just be confusing for ordinary users of learnr.

For this example, I think you also don't want `learnr::`, to keep things simple for users.

Here's what you'd end up with after those two changes. This will be easier for new users to comprehend:

```R
safe(run_tutorial("slidy", package = "learnr"))
```

schloerke: No problem, I can remove the brackets.  

To get the code to work, I'd need to library the learnr package.  Makes sense as it's in the learnr package.

schloerke:
@wch Any idea on how to get around not having to require the user to say 
```
safe(run_tutorial("slidly", package = "learnr", list(launch.browser = TRUE)))
```
as the `run_tutorial` is executed in a non interactive environment. 

Environment variable?

PR #175, opened 2018-07-30
Add `envir_prep` to checker function call

### LINKED ISSUE #163: Feature request: pass checker function the environment as it is _before_ student code is run
learnr passes the exercise checking function `envir_result`, which is the state of the R environment for the exercise **after** the chunk is executed ([as documented here](https://rstudio.github.io/learnr/exercises.html#exercise_checking)).

I could simplify checking if I had the R environment for the exercise as it is **before** the chunk is executed (but after any setup chunks are executed). This would let me:
1. Compare the student result to the solution result by running the solution code and the student code in the same environment
2. Diagnose where the student goes wrong, by running the student's call tree and the solution call tree step-by-step in the same environment
3. Ignore tracking down setup chunks for the exercises in the checking code (These need to be run before the student code, but I haven't found a good way to find them at check time)

schloerke:
Fixes #163

Submits a copy of the exercise environment before the submission is rendered.  This gives a fresh environment for the checker to user that contains the setup code.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -17,3 +17,12 @@ is_localhost <- function(location) {
   else
     FALSE
 }
+
+
+# Create an new, 'twin' environment with the same objects and same parent.
+twin_env <- function(envir, parent = parent.env(envir)) {
+  new_envir <- new.env(parent = parent)
+  for (object in ls(envir = envir))
+    new_envir[[object]] <- envir[[object]]
+  new_envir
+}
--- a/R/exercise.R
+++ b/R/exercise.R
@@ -52,14 +52,11 @@ setup_exercise_handler <- function(exercise_rx, session) {
       else
         evaluator_factory <- inline_evaluator
     }
-    
-    # create a new environment parented by the global environment 
-    envir <- new.env(parent = globalenv())
-    
+
+    # create a new environment parented by the global environment
     # transfer all of the objects in the server_envir (i.e. setup and data chunks)
-    for (object in ls(envir = server_envir))
-      envir[[object]] <- server_envir[[object]]
-    
+    envir <- twin_env(server_envir, parent = globalenv())
+
     # create exercise evaluator
     evaluator <- evaluator_factory(evaluate_exercise(exercise, envir), timelimit)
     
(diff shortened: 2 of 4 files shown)


jjallaire:
We try not to have so many spurious whitespace changes in a diff (makes it hard to tell what material code changes there are). It would be great if you could remove these.

If you feel strongly about enforcing a whitespace convention I'd suggest just one big commit that fixes all the whitespace (and updates the .Rproj file to have a setting that agrees with the scheme).

schloerke:
@jjallaire Thanks for the feedback! I've reduced the changes.  (I'll make a separate PR for .Rproj whitespace changes later.) . 

I dropped compiling within the `./docs` folder as it produces many changes.  Similar to shiny, I think we can wait until CRAN release.

- [ ] NEWS item
- [ ] test?

jjallaire:
Great, thanks!

R/utils.R
```diff
@@ -17,3 +17,12 @@ is_localhost <- function(location) {
   else
     FALSE
 }
+
+
+# Create an new, 'twin' environment with the same objects and same parent.
+twin_env <- function(envir, parent = parent.env(envir)) {
+  new_envir <- new.env(parent = parent)
+  for (object in ls(envir = envir))
```
wch: Should this use `all.names=TRUE`?

wch: Some thoughs that aren't super important, but they might be slightly useful because the function is exported....

It might be faster with this, though you should probably test. 

```R
list2env(as.list.environment(envir, all.names = TRUE, sorted = FALSE))
```

Some notes:

The `sorted=FALSE` is because sorting can take a nontrivial amount of time.

```R
system.time(ls(baseenv()))
#>   user  system elapsed 
#>  0.007   0.000   0.006 

system.time(ls(baseenv(), sorted=FALSE))
#>   user  system elapsed 
#>  0.003   0.000   0.003 
```

The reason it uses an explicit `as.list.environment` is because if the environment happens to have a class on it, it will error. For example:

```R
e <- new.env()
e$x <- 1
class(e) <- "foo"

as.list(e)
#> Error in as.vector(x, "list") : 
#>   cannot coerce type 'environment' to vector of type 'list'

as.list.environment(e)
#> $x
#> [1] 1
```

PR #179, opened 2018-08-10
Aggressive Prerender checks within rmarkdown

### LINKED ISSUE #169: error when running a tutorial
I am new to R. I am trying to follow a tutorial for R.

After I load the learnr package with library(learnr), I try to run a tutorial with learnr::run_tutorial("hello", package="learnr"), I get the following error in my console:

```
Loading required package: shiny

Listening on http://127.0.0.1:5950
Warning: Error in value[[3L]]: Couldn't normalize path in `addResourcePath`, with arguments: `prefix` = 'font-awesome-4.5.0'; `directoryPath` = ''
  [No stack trace available]
```

At the same time a page is opened on http://127.0.0.1:5950/ displaying the following error:

> ERROR: Couldn't normalize path in `addResourcePath`, with arguments: `prefix` = 'font-awesome-4.5.0'; `directoryPath` = ''

How can I fix this?

schloerke:
Fixes #169 

Uses rstudio/rmarkdown#1420 to rerender tutorials if any html dependency package version is different and if any package used to build the document is different at run time.  

Because this is so aggressive and the checks should be done every time a user runs the document, setting the environment variable `RMARKDOWN_RUN_PRERENDER=0` should not be done.


schloerke:
Will require a followup PR to remove the reference in the DESCRIPTION file once rmarkdown is released to CRAN

jjallaire:
One scenario we should be sure to encompass is to be able to run tutorials in an environment where they are *never* rendered (we need this e.g. for Shiny Server when they are deployed in a readonly directory). Is that still possible?

schloerke:
`rmarkdown` still has the hooks for skipping the pre-render.  

Can still execute this to never prerender:
```r
withr::with_envvar(c(RMARKDOWN_RUN_PRERENDER = "0"), {	    
  learnr::run_tutorial(FILE, package = PKG)
})
```

Would it be better to have a `run_tutorial_local` function that implements the _always possibly render_ functionality? (Keeping `run_tutorial` as is)
```r
learnr::run_tutorial_local(FILE, package = PKG)
```

jjallaire:
No, I think it's fine as long as rmarkdown has that hook.

NEWS.md
```diff
@@ -1,8 +1,11 @@
-learnr 0.9.3
+learnr 0.9.3 (unreleased)
```
wch: 0.9.2.9000

PR #194, opened 2018-10-05
Make shiny modules for quiz questions (instead of htmlwidgets)

### LINKED ISSUE #109: Fonts in different languages
I am trying to prepare R tutorials in my own language (Turkish). There are some letters in Turkish like: ş, ğ, ü, ö, ç, ı, Ş, İ, Ğ, Ü, Ö, Ç.

I couldn't find a way to print these letters in learnr. I was able to solve this issue in tufte package by using "main-font: UTF-8". Could it be possible for learnr also?

### LINKED ISSUE #173: non ascii characters are not rendered correctly in quizz
``````
---
title: "àéïôu"
output: learnr::tutorial
runtime: shiny_prerendered
---
  
```{r}
library(learnr)
```

àéïôu

```{r aeiou}
quiz(
  question(text="_à_",
           answer("_é_"),
           answer("_ï_"),
           answer("_ô_"),
           answer("u",correct=TRUE)
  )
)
```
``````

maybe related to [Fonts in different languages #109](https://github.com/rstudio/learnr/issues/109)

### LINKED ISSUE #192: Scrambled json
@garrettgman 

With the current htmlwidget implementation, the JSON text is provided in plain text.  This contains information of "This answer is the correct answer".  

Should this be scrambled / hidden from the users?

screen shot 2018-09-20 at 9 52 37 am

(I don't know of a direct way to do this without also exposing this to the users as well.)

### LINKED ISSUE #217: Returns to First Slide After "submit" in Slidy
I am trying to embed questions inside of a slidy presentation. 

Every time you answer a question, though, it sends you back to the first slide. Thoughts? 

[WomenMenQuestions_V3_slidy.zip](https://github.com/rstudio/learnr/files/2959455/WomenMenQuestions_V3_slidy.zip)

### PR DESCRIPTION
**Pull request by schloerke**
Fixes #192
Fixes #173
Fixes #109
Fixes #217

~Will make a separate PR for text box input-based questions~ (included)

### CODE REVIEW

**Comment by jcheng5** [comment]
1. Looks like there are some `TODO-barret` comments left, do any of those need to be addressed?
2. Are you aware of any backward-incompatible changes in this PR? I've only started skimming but I vaguely recall you saying you were getting rid of the `quiz` function?

**Comment by jcheng5** [comment]
To (partially) answer my own question, the `question()` arguments `correct` and `incorrect` have been renamed to `correct_message` and `incorrect_message`. Do you really want to break backward compat for that? (The examples still use the old param names btw)

**Comment by jcheng5** [comment]
I can't get this to work in RStudio. It seems to me like the S3 methods are not registered properly. This is what I'm testing with:

    ---
    title: "learnr test"
    output: learnr::tutorial
    runtime: shiny_prerendered
    ---
    
    ```{r setup, include=FALSE}
    library(learnr)
    ```
    
    ```{r one}
    question("Where are you right now? (select ALL that apply)",
     answer("Planet Earth", correct = TRUE),
     answer("Pluto"),
     answer("At a computing device", correct = TRUE),
     answer("In the Milky Way", correct = TRUE),
     incorrect_message = paste0("Incorrect. You're on Earth, ",
                        "in the Milky Way, at a computer.")
    )
    ```

**Comment by jcheng5** [comment]
Also there seem to be a lot of missing import declarations, R CMD check will point them out.

File: R/quiz.R [hunk 1]
```diff
@@ -563,86 +592,153 @@ question_module_server <- function(
...
+
   
-  message_container_info <- reactiveVal(NULL, label = "message container info")
-  output$message_container <- renderUI({
-    question_messages(question, req(message_container_info()))
+  button_type <- reactive(label = "button type", {
+    if (is.null(submitted_answer())) {
+      "submit"
+    } else {
+      req(!is.null(is_correct_info()))
```
jcheng5: If we ever get here and `is_correct_info()` is `NULL`, that's a programmer error somewhere, right? Because it means that `question_is_correct` returned `NULL`, which seems like it should never be the case?

If so, this feels more like an assertion than validation; you might want to make this a `stop` instead of a `req`.

R/quiz.R
```diff
@@ -563,86 +592,153 @@ question_module_server <- function(
...
+        } else {
+          # not correct and can not try again
+          "incorrect"
+        }
+      }
+    }
+  })
+  
+  # disable / enable for every input$answer change
+  answer_is_valid <- reactive(label = "answer_is_valid", {
```
jcheng5: What's the intended behavior here? Before you've submitted your first answer, you get realtime feedback, but after you've submitted then you don't?

R/quiz.R
```diff
@@ -563,86 +592,153 @@ question_module_server <- function(
...
+    } else {
+      question_is_valid(question, submitted_answer())
+    }
   })
   
 
-  init_question <- function() {
+  init_question <- function(restoreValue = NULL) {
     if (question$random_answer_order) {
       question$answers <<- shuffle(question$answers)
```
jcheng5: Hmmm, this `<<-` side effect feels slightly out of place. Although I can't say I have a suggestion for how to factor it... more of a code smell.

R/quiz.R
```diff
@@ -563,86 +592,153 @@ question_module_server <- function(
...
     }
-    answer_container(question_initialize_input(question, NULL))
-    message_container_info(NULL)
-    button_type("submit")
+    submitted_answer(restoreValue)
+  }
+  
+  # TODO-barret get storage state
+  
+  past_submission <- retrieve_question_submission(session, question$label)
```
jcheng5: Leaving off here for lunch. bbiab

R/quiz.R
```diff
@@ -538,6 +546,27 @@ question_completed_input.text <- function(question, answer_input, ...) {
...
+
+  # create submissions from state objects
+  submissions <- submissions_from_state_objects(state_objects)
+  
+  submissions
+}
+
+retrieve_question_submission <- function(session, question_label) {
+  question_label <- as.character(question_label)
+  submissions <-
```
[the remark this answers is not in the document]
schloerke: Yup. Removed

R/quiz.R
schloerke: Correct! Made into a `stop()`

R/quiz.R
schloerke: Correct.  I'm trying to safeguard against users completely wiping the answer area, such as returning "You've submitted an answer!" (and therefore `input$answer` does not map to anything existing in the UI).  

By keeping the submitted answer, the logic should stay valid whether or not the completed answer UI supports `input$answer`.

Question: If the answer UI doesn't support `input$answer`, does `input$answer` continue to exist?  Feels dirty if this true.

R/quiz.R
schloerke: The goal is to change the answer order for the `question` variable within `question_module_server_impl`.  This allows for the `question_is_valid` and `question_is_correct` S3 methods to receive a question object that matches exactly what the user sees.  

The logic could be done locally if matching answer order within the answer checking functions is not important. (Always submitting the original question object).

R/quiz.R
schloerke: 🍔🌯🍦

PR #219, opened 2019-04-04
Add tabset support

### LINKED ISSUE #212: Create dynamic tabset
Within a standard rmarkdown document, we can create tabsets by using the html syntaxt `{.tabset}` after a title. Example:

```
---
title: "Markdown"
output:
  html_document:
    self_contained: true
---

## Introduction {.tabset}    
### Base R
This is the first tabset
### Tidyverse
This is a second tabset
```

But when we use `output: learnr::tutorial:`, the dynamic tabsets are not working any more, they appear as subtitles. 
```
---
title: "Markdown"
output:
  learnr::tutorial:
    self_contained: true
runtime: shiny_prerendered
---
```

Is there a way to make dynamic tabsets working ?

### LINKED ISSUE #213: Feature Request: Enable non-R code exercises
In normal Rmd documents you can specify SQL-only chunks, based on a database connection via DBI:

````
```{r}
library(DBI)
mydata = dbConnect(RPostgres::Postgres(), 
               dbname = "some_database",
               host = 'localhost',
               port = 5432,
               user = 'postgres',
               password = 'mysecretpassword')
```
````

````
```{sql, connection=mydata}
CREATE TABLE table_name (
  PersonID int,
  LastName varchar(255),
  FirstName varchar(255)
);
```
````

However, this does not work with `learnr`. I would like to be able to do the same in an exercise so that students can directly submit SQL statements:

````
```{sql, connection=mydata, exercise = TRUE}

```
````

Where students then would have to enter SQL code into the exercise.

schloerke:
Fixes #212

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 learnr 0.9.2.9000 (unreleased)
 ===========
 
+* Added tabset support. ([#219](https://github.com/rstudio/learnr/pull/219) [#213](https://github.com/rstudio/learnr/issues/213))
+
 * Fixed a spurious console warning when running exercises using Pandoc 2.0. ([#154](https://github.com/rstudio/learnr/issues/154))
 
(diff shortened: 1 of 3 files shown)


jcheng5:
Can you add a comment to the HTML indicating where the CSS came from? Other than that, LGTM.

PR #228, opened 2019-04-25
Retrieve last value from evaluating exercises.  Pass this to `checker` functions

### LINKED ISSUE #198: use result of last statement in exercise.checker function
Hi,

A natural thing to do in R is to type a statement in the console and inspect the result. If a reader does this in an exercise I would like to use the result of the last statement in an exercise.checker function. Since it's nameless, it's not in envir_result. Evaluate_result contains a representation of the last statement but not the value of the R object. I currently use `try(eval(user_code))` but that slows things down since the same code is now evaluated twice.

Is there a better option to get at this value? If not, could it be added?

Thanks!

schloerke:
Fixes #198


schloerke:
small round trip test with grader works

R/exercise.R
```diff
@@ -197,9 +198,37 @@ evaluate_exercise <- function(exercise, envir) {
...
+  learnr_output_handler <- evaluate::new_output_handler(value = function(x, visible) {
+    last_value <<- x
+
+    if (has_visible) {
+      default_output_handler$value(x, visible)
+    } else {
+      if (visible) {
+        default_output_handler$value(x)
+      } else {
+        NULL
```
wch: Should this return `list(value = NULL, visible = FALSE)`?  See https://github.com/r-lib/evaluate/blob/8049bb53e7c1cc7ed2ca4adc428145a6f715ca69/R/eval.r#L214

wch: Maybe this should be `invisible()`.

PR #242, opened 2019-06-07
do not stop when an unknown tag type is found in mutate_tags

schloerke:
PR task list:

Helpful when making `sortable` question extension and running into a `sortable` S3 object.  It has not other types and should just be inspected further.

Currently, `learnr` throws when it reaches the `sortable` object when trying to `mutate_tags`.  `learnr` should not do this.  If users will want to have a custom `mutate_tags` method, they are still able to write `mutate_tags.TYPE` methods themselves (so we are not losing any functionality, only making basic functionality a smoother experience).


jcheng5:
Consider calling `as.tags(ele)`

PR #248, opened 2019-07-01
Add `question_try_again_input`, export `disable_all_tags`, add `question$seed`

schloerke:
Fixes #244 
Fixes #246 

* `mutate_tags` uses an lapply to avoid setting a `NULL` value which would remove an element in a for-loop
* Added a random seed in the question object for questions to use per user session
* export `disable_all_tags`
* added method of `question_try_again_input`
* `question_completed_input` defaults to disabling all tags of `question_initialize_input`

PR task list:
- [ ] Update NEWS
- [ ] Add tests (if possible)

cc @andrie


R/question_methods.R
```diff
@@ -56,8 +61,10 @@ question_stop <- function(name, question) {
 question_initialize_input.default <- function(question, answer_input, ...) {
   question_stop("question_initialize_input", question)
 }
-question_completed_input.default <- function(question, ...) {
-  question_stop("question_completed_input", question)
+question_completed_input.default <- function(question, answer_input, ...) {
+  disable_all_tags(
+    question_initialize_input(question, answer_input)
```
wch: Use `...`

R/question_methods.R
```diff
@@ -66,6 +73,12 @@ question_is_correct.default <- function(question, answer_input, ...) {
   question_stop("question_is_correct", question)
 }
 
+question_try_again_input.default <- function(question, answer_input, ...) {
+  disable_all_tags(
+    question_initialize_input(question, answer_input)
```
wch: Use `...`

PR #259, opened 2019-07-17
install_tutorial_dependencies()

### LINKED ISSUE #253: Feature request: ask to install tutorial dependencies
Hi! @jules32, @angela-li, and I are working on improving the `use_tutorial()` function for `usethis`, which sets up the `learnr` tutorial package format . Our two main use cases are (1) package developers that want to include tutorials as documentation and (2) course instructors that want to distribute a bundle of tutorials to students.

For course instructors, it's especially important to minimize friction related to package installs. Thus, it would be great if `run_tutorial()` would prompt the user to install packages that the tutorial depends on.  

@cderv recommended 3 options to find the dependencies in the tutorial:

- [`renv::dependencies()`](https://rstudio.github.io/renv/reference/dependencies.html) - returns a list of dependencies contained in a file/folder
- [`attachment::att_from_rmds()`](https://thinkr-open.github.io/attachment/) - returns a list of dependencies contained in a file/folder
- [`rsconnect::appDependencies()`](https://rdrr.io/cran/rsconnect/man/appDependencies.html) - returns a *full tree* of dependencies contained in a file/folder

`utils::menu()` could then be used to ask users if it's okay to install all the dependencies.

isteves:
Fixes #253 

Hi @schloerke, I've drafted a PR for your review based on our discussion in #253.

Happy for any/all feedback on how it's written and how I've tested it.  I wasn't able to get rid of all the NOTEs/WARNINGs, but looks like some of them come from other files.  Also, I attempted to use `with_mock()` in testing but couldn't figure out how to suppress the `menu()`.  (I've commented it out for now until it gets solved or deleted.)

PR task list:
- [ ] Update NEWS
- [x] Add tests (if possible)

### DIFF
--- a/R/install_tutorial_dependencies.R
+++ b/R/install_tutorial_dependencies.R
@@ -1,0 +1,30 @@
+install_tutorial_dependencies <- function(dir) {
+  file_paths <- list.files(dir,
+                           pattern = "[.]R$|[.]Rmd",
+                           full.names = TRUE,
+                           recursive = TRUE)
+  deps <- lapply(file_paths, packrat:::fileDependencies)
+  deps <- unique(unlist(deps))
+
+  need_install <- deps[!deps %in% utils::installed.packages()]
+
+  if(length(need_install) < 1) {
+    return(invisible())
+  }
+
+  need_install_formatted <- paste("  -", need_install, collapse = "\n")
+  question <- sprintf("Would you like to install the following packages?\n%s",
+                      need_install_formatted)
+
+  if(!interactive()) {
+    stop("The following packages need to be installed:\n",
+         need_install_formatted)
+  }
+
+  answer <- utils::menu(choices = c("yes", "no"),
+                        title = question)
+
+  if(answer == 2) stop("The tutorial is missing required packages and cannot be rendered.")
+
+  utils::install.packages(need_install)
+}
--- a/R/install_tutorial_dependencies.R
+++ b/R/install_tutorial_dependencies.R
@@ -1,30 +1,43 @@
-install_tutorial_dependencies <- function(dir) {
+get_needed_pkgs <- function(dir) {
   file_paths <- list.files(dir,
                            pattern = "[.]R$|[.]Rmd",
                            full.names = TRUE,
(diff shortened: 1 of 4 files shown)


R/utils.R
```diff
@@ -1,5 +1,7 @@
 "%||%" <- function(x, y) if (is.null(x)) y else x
 
+fileDependencies <- getFromNamespace("fileDependencies", "packrat")
```
schloerke: change to `packrat_file_dependencies`, please

R/install_tutorial_dependencies.R
```diff
@@ -0,0 +1,43 @@
+get_needed_pkgs <- function(dir) {
```
schloerke: instead of using `fileDependencies`, we could maybe use `pkgs <- packrat:::dirDependencies(dir)`

R/run.R
```diff
@@ -39,6 +38,8 @@ run_tutorial <- function(name = NULL, package = NULL, shiny_args = NULL) {
   # get path to tutorial
   tutorial_path <- get_tutorial_path(name, package)
 
+  install_tutorial_dependencies(tutorial_path)
```
schloerke: add comment of `"# check for necessary tutorial package dependencies`

R/utils.R
schloerke: (or `packrat_dir_dependencies`)

schloerke: Thinking about this some more, let's move this definition inside `get_needed_pkgs` to avoid package issues.

(on demand vs statically compiled)

If done inside `utils.R`, the function value will be statically compiled at the build time of `learnr`.  If packrat updates the internals of their function and a new packrat is installed, `learnr` will not pick up the changes as it was compiled with the first version definition.

If it is retrieved within the function, it will always have the latest.

schloerke: If cran/check gets mad about getFromNamespace, we can use `packrat:::yoink` definition

```r
eval(call(":::", pkg, func))
```

schloerke:
Please add a news entry for this enhancement pointing to this PR (or the issue) with your handle tagged.

PR #281, opened 2019-09-18
Enable Mathjax in quiz questions

schloerke:
If text is formatted for mathjax, it is never triggered to be rendered.

```r
run_tutorial("quiz_question", "learnr")
# Look at `Formatting and Math` section in `Overview`
```

PR task list:
- [NA] Update NEWS

### DIFF
--- a/R/install_tutorial_dependencies.R
+++ b/R/install_tutorial_dependencies.R
@@ -2,6 +2,9 @@ get_needed_pkgs <- function(dir) {
 
   pkgs <- unique(renv::dependencies(dir, quiet = TRUE)$Package)
 
+  # remove packages with name "cannot open connection"
+  pkgs <- pkgs[!grepl(" ", pkgs)]
+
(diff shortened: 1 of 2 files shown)


R/install_tutorial_dependencies.R
```diff
@@ -2,6 +2,9 @@ get_needed_pkgs <- function(dir) {
 
   pkgs <- unique(renv::dependencies(dir, quiet = TRUE)$Package)
 
+  # remove packages with name "cannot open connection"
```
wch: File an issue about this bug in renv, and refer to it here in the comment.

schloerke: moved code to https://github.com/rstudio/learnr/pull/292

schloerke:
Have restructured the PR to call a JS method to run mathjax.  Mathjax is included by default in the rmarkdown document (even if mathjax is not required).  If mathjax isn't found, nothing is triggered.  If mathjax isn't found, this will be because the user has removed it.

PR #291, opened 2019-10-28
Use a version for `question_submission` events

schloerke:
Follow up to https://github.com/rstudio/learnr/pull/276

PR task list:
- [x] Update NEWS


R/storage.R
```diff
@@ -205,18 +195,41 @@ save_object <- function(session, object_id, data) {
...
 }
 
+
+update_object <- function(object) {
+  if (object$type == "question_submission") {
+    version <- object$data$version
+    if (!is.null(version)) {
+      if (identical(version, 1)) {
+        # do nothing
+      } else {
```
wch: This `else` should be for `!is.null(version)`, right?

R/storage.R
```diff
@@ -5,6 +5,7 @@ save_question_submission <- function(session, label, question, answer) {
     session = session,
     object_id = label,
     tutorial_object("question_submission", list(
+      version = 1,
```
wch: How about `api_version`, so that authors don't get confused and think this is a field that they can/should change?

PR #305, opened 2019-11-06
Use indexedDB directly via idb-keyval

### LINKED ISSUE #304: Quiz questions are not loading when using multiple tabs
reprex:
* Open many tutorials on shinyapps.io on different tabs.
* The first tutorial will work fine. Other tutorials will have problems advancing or saving progress (or any tutorial state) to the browser.

The tutorial tabs are all using `localforage` to access the `indexeddb` in the web browser.  This DB only allows for a single connection per browser.  Once the tab closes, that connection is released and the remaining tabs attempt to connect.

There is no warning or error.  

This problem exists in v0.9.2, but the quiz questions would display before any answers were being restored.  We could do this, but opted not to, as changing the UI after some non trivial time has passed is confusing.  (The downside is that some tabs can not display their quiz questions.)

Original note: https://github.com/rstudio/learnr/blob/fe4fe442850324dc17ef8aa9d645396ad8a2fbea/inst/lib/tutorial/tutorial.js#L1454-L1457

With `indexeddb` being [widely supported](https://caniuse.com/#search=indexeddb) (95%+), it could be used natively and `localforage` could be dropped as our requirements are slim (essentially key/value lookup).  `localstorage` could also be used instead, but there is a limitation of 5mb per domain.  `indexeddb` would be best going forward.

Known issues:
* https://stackoverflow.com/questions/51615658/using-indexeddb-operation-in-multiple-tabs
* https://github.com/localForage/localForage/issues/912
* https://github.com/localForage/localForage/issues/179
  * using `this.localforage._dbInfo.db.close()` does close the connection. However, the connection was not allowed to reconnect. 😞

schloerke:
Fixes #304
Fixes #302

(View diff with white space turned off.)

Notes: 
https://github.com/rstudio/learnr/pull/305/files?utf8=✓&diff=split&w=1#diff-c7d58e72ba1e4c97d7f872330fd63d87R1473-R1494

PR task list:
- [x] Update NEWS
- [x] Update authors in DESCRIPTION

Includes code from https://github.com/jakearchibald/idb-keyval/pull/73/files#diff-32d9810b60fe9355ae31121d8cc181e2


NEWS.md
```diff
@@ -25,6 +25,7 @@ learnr 0.10.0 (unreleased)
 
 * A version number has been added to `question_submission` events.  This will help when using custom storage methods. ([#291](https://github.com/rstudio/learnr/pull/291))
 
+* Tutorial storage on the browser is now executed directly on `indexedDB` using `idb-keyval` (dropping `localforage`). ([#305](https://github.com/rstudio/learnr/pull/305))
```
wch: I think it's worth mentioning that multiple tabs won't block each other.

inst/lib/tutorial/tutorial.js
```diff
@@ -1444,106 +1444,185 @@ Tutorial.prototype.$clearExerciseOutput = function(exercise) {
...
-    }).then(function() {
-      success(objects);
-    }).catch(function(err) {
-      console.log(err);
-      success(objects);
-    });
-  } else {
-    success(objects);
+
+  // overwrite prototype to clear out all key/vals
```
wch: "masking" is a bit clearer

inst/lib/tutorial/tutorial.js
```diff
@@ -1444,106 +1444,185 @@ Tutorial.prototype.$clearExerciseOutput = function(exercise) {
...
+      return Promise.all(getPromises)
+        .then(function(vals) {
+          var ret = {}, i;
+          for (i = 0; i < keys.length; i++) {
+            ret[keys[i]] = vals[i];
+          }
+          return ret;
+        });
+    })
+    .then(function (objs) {
```
wch: Extra space

wch:
We should also include license information for idb-keyval.

PR #309, opened 2019-11-27
Python code chunk support within learnr

cassws:
(This is an early PR as a place to discuss these changes, but please let me know if another format for discussing/troubleshooting a PR is better!)

As per #213 I am working on a PR to enable Python (and eventually more language) support in learnr. As proof of concept, I was able to modify exercise.R to set `knitr::opts_chunk$set(engine='python')` and test with code permissible in syntax in R and Python, but with different evaluation:

```
x = 8
x <-10
print(x)
```
Now returns:
`False`
`8`

Indicating running Python engine. However, I was unable to set this on a chunk by chunk basis despite attempting to modify knitr-hooks.R to pass along the engine name (e.g. adding to preserved_options to pass the`python` in `{python exercise=TRUE...}` along as `options$engine`).

**Question**: is there another strategy for passing along and accessing the engine option for a code chunk within exercise.R? I feel like I'm missing an aspect of the interaction between code chunks -> knitr -> exercise.R.

I'll be able to move on to working with code checking/etc. issues as per #213 once I can access the engine parameter from within exercise.R. Thank you!


zoews:
Oh no! My apologies. I tried to resolve an authorship issue with my two Github accounts and introduced a bunch of unnecessary historical commits. I will delete and try again.

Issue #312, opened 2019-12-16
tag learnr tutorials with extra metadata?


kevinushey:
As part of the Tutorial pane work in the RStudio IDE, I'd like to make a landing page for all discovered tutorials provided by any installed packages. It currently looks like this:

![Screen Shot 2019-12-16 at 3 10 18 PM](https://user-images.githubusercontent.com/1976582/70950874-b23e0600-2016-11ea-94a2-a3a695fe99ea.png)

The presentation is currently a bit bare. I wonder if there is some other relevant metadata we could include for a particular tutorial, e.g.

- A short, single-paragraph description? [description]
- The R package(s), if any, one might learn about in this tutorial? [packages]

The format I'm currently imagining would be something like:

> Data Basics [ex-data-basics]
> 
> Learn about the base data types in R. Explore R's data frames, and learn how to interact with data frames and their columns using dplyr.
> 
> Packages: dplyr, tibble

All the relevant data here could be included as part of the YAML header.

coatless:
@kevinushey any way to render a thumbnail of the tutorial in the pane?

kevinushey:
Not currently, but we could allow for something like `icon.png` if that would be helpful.

coatless:
@kevinushey that would be great. Though, might need to enforce image size (storage + display).

PR #313, opened 2019-12-16
add descriptions to bundled tutorials; document use

kevinushey:
Closes https://github.com/rstudio/learnr/issues/312.


kevinushey:
Here's the current treatment I have for this PR:

Screen Shot 2019-12-18 at 1 21 44 PM

I'd like to propose (if this PR is accepted) that new tutorials should have a short one-paragraph description going forward.

I'd also like to propose the use of an `internal` field to indicate tutorials that might not want to be exposed / indexed by front-ends (ie, they exist just as example tutorials for developers)

schloerke:
Do you think we could use `private` instead of `internal`?

Reminds me of https://docs.npmjs.com/files/package.json#private

kevinushey:
Sounds good to me!

schloerke:
@kevinushey 

Would you like this information (and possibly the whole yaml header information) to be supplied in `learnr::available_tutorials()`? 

Is there a way we can move the template from within RStudio to within learnr? So as more information becomes available, the display is tied to learnr itself.  ... Is this a wrong approach?

kevinushey:
> Would you like this information (and possibly the whole yaml header information) to be supplied in learnr::available_tutorials()?

Yes, that would be useful.

> Is there a way we can move the template from within RStudio to within learnr? So as more information becomes available, the display is tied to learnr itself. ... Is this a wrong approach?

The main reason I decided to implement this first in RStudio is I wanted to make sure the generated HTML could be easily styled to accommodate both light + dark RStudio themes. I'd be open to having a solution in `learnr`, though.

schloerke:
With the same logic... I'm going to not display the `private: true` tutorials in `available_tutorials()`

PR #328, opened 2020-02-11
Add session start/stop events

trestletech:
Adds `session_start` and `session_stop` events.

I originally hoped to install the session_start handler in `initialize.R` next to the stop handler. Unfortunately, we don't `initialize_session_state` (which is necessary to get the tutorial ID/name/etc) until we run a `session$registerDataObj` handler. I couldn't find a viable way to defer the session start logic until the handler had been run, so I went this route instead. (`later::later` didn't do it late enough).

```
Listening on http://127.0.0.1:6924
com.rstudio.primers.02-vis-basics (1): jeff
event: session_start
com.rstudio.primers.02-vis-basics (1): jeff
event: section_skipped
com.rstudio.primers.02-vis-basics (1): jeff
event: question_submission
com.rstudio.primers.02-vis-basics (1): jeff
event: session_stop
```

PR task list:
- [x] Update NEWS
- NO - Add tests (if possible) -- deemed unfeasible?
- N/A - Update documentation with `devtools::document()`

### DIFF
--- a/R/initialize.R
+++ b/R/initialize.R
@@ -33,6 +33,15 @@ initialize_tutorial <- function() {
       singleton = TRUE
     )
 
+    # Register session stop handler
+    rmarkdown::shiny_prerendered_chunk(
+      'server',
+      sprintf('session$onSessionEnded(function() {
+        learnr:::record_event(session = session, event = "session_stop", data = list())
+      })'),
+      singleton = TRUE
+    )
+
     # set initialized flag to ensure single initialization
     knitr::opts_knit$set(tutorial.initialized = TRUE)
   }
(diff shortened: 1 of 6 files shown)


R/initialize.R
```diff
@@ -33,6 +33,15 @@ initialize_tutorial <- function() {
       singleton = TRUE
     )
 
+    # Register session stop handler
+    rmarkdown::shiny_prerendered_chunk(
+      'server',
+      sprintf('session$onSessionEnded(function() {
+        learnr:::record_event(session = session, event = "session_stop", data = list())
```
schloerke: Cosmetic change... Can we call something like `learnr:::session_stopped_event(session)` instead? 

And define both `session_stopped_event` and `session_started_event` in `event.R`?

trestletech: Done!

PR #329, opened 2020-02-11
implement tutorial_package_dependencies()

kevinushey:
For front-ends that might want to enumerate a tutorial's required R package dependencies.

This is admittedly quick-and-dirty; @schloerke let me know if this looks okay or if you'd prefer some tweaks.


NAMESPACE
```diff
@@ -40,6 +40,7 @@ export(run_tutorial)
 export(safe)
 export(safe_env)
 export(tutorial)
+export(tutorial_dependencies)
```
kevinushey: We could also call this `tutorial_package_dependencies()` if you think enumerating dependencies that aren't R packages would ever become important.

schloerke: I like safeguarding...

Starting to get a long name... `tutorial_pkg_deps(name, package)`?  Or experience says keep the long/clear name?

kevinushey: Your call! I usually rely on autocomplete to fill in longer names so those don't bother me as much; especially for a rarely-used API. But it's personal preference.

schloerke:
Two major changes.

* Adds column `package_dependencies` to `available_tutorials()` output
* Adds function `tutorial_package_dependencies()`

NEWS.md
```diff
@@ -5,12 +5,16 @@ learnr 0.10.0.9000 (unreleased)
 
 ## Minor new features and improvements
 
+* `learnr` gained the function `learnr::tutorial_dependencies()`, used to enumerate a tutorial's R package dependencies. Front-ends can use this to ensure a tutorial's dependencies are satisfied before attempting to run that tutorial.
```
wch: Would be good to have a link to this PR.

R/tutorial_package_dependencies.R
```diff
@@ -0,0 +1,72 @@
...
+}
+
+
+
+
+#' List Tutorial Dependencies
+#'
+#' List the \R packages required to run a particular tutorial.
+#'
+#' @param name The tutorial name. If \code{name} is \code{NULL}, then all tutorials within \code{package} will be searched.
```
wch: Tiny nit: Can you wrap this text using the RStudio IDE?

R/tutorial_package_dependencies.R
```diff
@@ -0,0 +1,72 @@
...
+#' @param package The \R package providing the tutorial. If \code{package} is \code{NULL}, then all tutorials will be searched.
+#'
+#' @export
+#' @return A character vector of package names that are require for execution.
```
wch: "required"

tests/testthat/test-dependency.R
```diff
@@ -6,3 +6,13 @@ test_that("tutor html dependencies can be retreived", {
   expect_equal(dep$name, "tutorial")
 })
 
+test_that("tutorial package dependencies can be enumerated", {
+  testthat::skip_if_not_installed("learnr")
```
wch: I think this will never cause it to skip.

PR #337, opened 2020-03-04
Separate exercise events, add additional metadata

trestletech:
* Renamed the `exercise_submission` event to `exercise_result` and added the following fields: 
  1. `id` - a randombly generated identifier that can be used to align with the associated `exercise_result` event.
  2. `time_elapsed` - the time required to run the exercise (in seconds)
  3. `timeout_exceeded` - indicates whether the exercise was interrupted due to an exceeded timeout. May be `NA` for some platforms/evaluators if that information is not known or reported.

* Added an `exercise_submitted` event which is fired before evaluating an exercise. This event can be associated with an `exercise_result` event using the randomly generated `id` included in the data of both events.


R/exercise.R
```diff
@@ -60,6 +60,19 @@ setup_exercise_handler <- function(exercise_rx, session) {
     # create exercise evaluator
     evaluator <- evaluator_factory(evaluate_exercise(exercise, envir), timelimit)
 
+    # Create exercise ID to map the associated events.
+    ex_id <- shiny:::createUniqueId(8)
```
trestletech: This generates a NOTE in R CMD check. I had originally just done a `runif` but I noticed all the magic that shiny does to isolate itself from external interactions with the random seed.

Should we just do a runif and call it a day, or copy in shiny's function?

schloerke: Could use `random_id("lnr_ex")`

```r
random_id("lnr_ex")
#> [1] "lnr_ex_bc1faf5"
```

https://github.com/rstudio/learnr/blob/0faa1ad9f792a0385111517755fa5562f40b1028/R/quiz.R#L247-L256

trestletech: Thanks!

PR #345, opened 2020-03-31
Add support for a remote evaluator

trestletech:
**Note that this PR is based on the `save-setup` branch, not master. So we'll either need to be considerate of the merge order or rebase this one once #346 lands on master.**

This change adds a new evaluator that supports running exercises remotely. It presumes a remote HTTP server that can receive R exercises that returns appropriately formatted responses.

This breaks expectations a bit, in that our current evaluators always inherit state from the parent RMD doc. Meaning that if you run `library()` somewhere in the initialization of your learnr doc, that package would already be loaded when your user goes to run an exercise. That will no longer be the case in this model. Instead, we rely on explicit use of the [`setup` chunks](https://rstudio.github.io/learnr/exercises.html#Exercise_Setup) for each exercise which must be used in order to provision context for an exercise.

## Changes

- Added a remote evaluator. This can be opted into by either setting the `tutorial.remote.host` option, setting the `TUTORIAL_REMOTE_EVALUATOR_HOST` environment variable, or using `new_remote_evaluator()` and assigning the result to the `tutorial.exercise.evaluator` option. The last option gives you the most control about how the remote evaluator is constructed.
- Adds two new arguments that get passed into the evaluator constructor; suggests that new evaluators accept `...` for future-proofing.
- Allows `evaluate-exercise` to opt-in to running the global setup chunk that's attached to the exercise prior to running exercise checkers. This allows remote sessions to have calls like `library(gradethis)` in the global setup and then leverage `gradethis` functions inside of their checkers.
- Updates docs and adds the OpenAPI schema for the remote evaluator.
- Adds an example called `evaluator-tests.Rmd` that exercises all the different configurations that I had in mind when working on this feature. This points to a server running on localhost; if you want to run against a remote server, ask me and I can send you the URL privately (avoiding posting the server publicly for now.)
- adds tests for the remote evaluator that runs an httpuv server in the background to more easily mock the responses and then measures the interactions the remote evaluator has with the server.

## Testing & Validation

Ask me for the URL of the server that can be used for testing.

Install this branch, if it hasn't already been merged `remotes::install_github("rstudio/learnr@remote")`

I think the bulk of the testing is just going to be trying a variety of inputs on the exercises with different configurations -- with and without setup chunks, with different time limits, and with various errors like invalid syntax, referencing variables that don't exist, etc. I've tried to capture all of the scenarios that I had in mind and that I tested in an RMD which I've included in this PR (`evaluator-tests.Rmd`). If you just adjust the host to point to the remote server, you should be able to just run through those and confirm that the output is what you expected. It might be interesting to compare the behavior/output of that RMD when running locally (i.e. comment out the `options(tutorial.remote.host` line at the top of the test RMD) versus running remotely. The only known difference I've seen is that invalid syntax has a bit more text in the error message: `Error in parse(text = x, keep.source = TRUE): <text>:4:8: unexpected input 3: 4: asdf + _ ^` vs `<text>:4:8: unexpected input 3: 4: asdf + _ ^`.

I don't expect much variation on different browsers, as there's not any new front-end code here. I'd like to test on different versions of R, but unfortunately the remote evaluator currently only has one version of R, so we're only going to be able to change the version of R on the learnr side, not the evaluation side. Probably still worth testing, though.


R/evaluators.R
```diff
@@ -118,3 +118,173 @@ forked_evaluator <- function(expr, timelimit) {
...
 }
+
+#' Remote execution evaluator
+#'
+#' @param endpoint The HTTP(S) endpoint to POST the exercises to
+#' @param max_curl_conns The maximum number of simultaneous HTTP requests to the
+#'   endpoint.
+#' @import curl
+#' @export
+new_remote_evaluator <- function(
```
schloerke, line 129: I'm worried about the name confusing about a prior remote evaluator.

* `remote_evaluator`?
* `external_evaluator`?
* `hosted_evaluator`?

schloerke: I'm partial to `hosted_evaluator`.  Gives me an indication that it's somewhere else.

R/evaluators.R
```diff
@@ -118,3 +118,173 @@ forked_evaluator <- function(expr, timelimit) {
...
+  endpoint = getOption("tutorial.remote.host", Sys.getenv("TUTORIAL_REMOTE_EVALUATOR_HOST", NA)),
+  max_curl_conns = 50){
+
+  internal_new_remote_evaluator(endpoint, max_curl_conns)
+}
+
+# An internal version of new_remote_evaluator that allows us to stub some calls
+# for testing.
+internal_new_remote_evaluator <- function(
```
schloerke, line 138: Is there a need for the separation of these two functions?

R/evaluators.R
```diff
@@ -118,3 +118,173 @@ forked_evaluator <- function(expr, timelimit) {
...
+            timeout_s <- 30 * 1000
+          } else {
+            timeout_s <- exercise$options$exercise.timelimit * 1000
+          }
+
+          # Create curl request
+          handle <- curl::new_handle(customrequest = "POST",
+                                     postfields = json,
+                                     postfieldsize = nchar(json),
+                                     timeout_ms = timeout_s + 5000)
```
schloerke, line 177:
```suggestion
                                     # add 5 seconds for application startup
                                     timeout_ms = timeout_s + 5000)
```

docs/openapi/openapi.json
```diff
@@ -0,0 +1 @@
...:"RemoteEvaluatorEvaluateExerciseResponseBody","type":"object","properties":{"error_message":{"type":"string","description":"If an error occurred, the plain-text representation","example":"object 'x' not found"},"feedback":{"type":"string","description":"TODO","example":"Consequatur minus sunt placeat eaque amet."},"html_output":{"type":"string","description":"The output of the given command","example":"\n\n\n\u003cpre\u003e\u003ccode\u003e[1] -0.78207928  0.70263075 -0.08438296 -0.06020523  1.97638361\u003c/code\u003e\u003c/pre\u003e\n"},"timeout_exceeded":{"type":"boolean","description":"True if the time limit was exceeded; otherwise false","example":false}},"example":{"error_message":"object 'x' not found","feedback":"Quo quisquam sit beatae corporis.","html_output":"\n\n\n\u003cpre\u003e\u003ccode\u003e[1] -0.78207928  0.70263075 -0.08438296 -0.06020523  1.97638361\u003c/code\u003e\u003c/pre\u003e\n","timeout_exceeded":true}},"RemoteEvaluatorInitiateSessionResponseBody":{"title":"RemoteEvaluatorInitiateSessionResponseBody","type":"object","properties":{"id":{"type":"string","description":"The ID of the session generated","example":"8rmci3kdkc"}},"example":{"id":"8rmci3kdkc"}}}}
```
schloerke, line 1: Are both files needed? (Prefer the `.yaml` file)

tests/testthat/test-evaluators.R
```diff
@@ -0,0 +1,315 @@
...
+    done <<- TRUE
+  }
+  err_cb <- function(res){
+    done <<- TRUE
+  }
+
+  initiate_remote_session(pool, paste0(srv$url, "/learnr/"), cb, err_cb)
+
+  while(!done){
+    later::run_now()
```
schloerke, line 115: I don't know if this will cause issues with `httpuv`. It should be ok, but if not, we could use `callr::r_bg` to launch the server.

docs/openapi/openapi.json
trestletech: I was just offering both for user convenience. But we could certainly kill the JSON if you'd prefer.

tests/testthat/test-evaluators.R
trestletech: Fair. So far I haven't had any issues while testing but we can keep an eye on it!

R/evaluators.R
trestletech: It's just used to mock stuff for testing. I tried to explain in the comment above but happy to rephrase if that makes it more clear. Or if the tests aren't work it, we can just kill them and make it all one function again.

schloerke: I missed the `initiate` variable.  🤦

docs/openapi/openapi.json
schloerke: One file helps avoid confusion down the road. (I would forget to update one or the other)

schloerke: Let's delete the JSON?

PR #353, opened 2020-04-20
Limit concurrent forks

trestletech:
Add a limit to the "forked" evaluator (the default used on Linux) to limit the number of forked exercises that learnr will execute in parallel. Previously, this was uncapped, which could cause a learnr process to run out of memory when an influx of traffic arrived. The default limit is 3, but it can be configured using the `tutorial.max.forked.procs` option or the `TUTORIAL_MAX_FORKED_PROCS` environment variable.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,17 +6,17 @@ learnr (development version)
 * Renamed the `exercise_submission` event to `exercise_result` and added the following fields: 
   1. `id` - a randombly generated identifier that can be used to align with the associated `exercise_result` event.
   2. `time_elapsed` - the time required to run the exercise (in seconds)
-  3. `timeout_exceeded` - indicates whether the exercise was interrupted due to an exceeded timeout. May be `NA` for some platforms/evaluators if that information is not known or reported.
+  3. `timeout_exceeded` - indicates whether the exercise was interrupted due to an exceeded timeout. May be `NA` for some platforms/evaluators if that information is not known or reported. ([#337](https://github.com/rstudio/learnr/pull/337))
   
 
 ## New features
 
-* Introduced an [experimental](https://www.tidyverse.org/lifecycle/#experimental) function `external_evaluator()` which can be used to define an exercise evaluator that runs on a remote server and is invoked via HTTP. This allows all exercise execution to be performed outside of the Shiny process hosting the learnr document.
-* For the "forked" evaluator (the default used on Linux), add a limit to the number of forked exercises that learnr will execute in parallel. Previously, this was uncapped, which could cause a learnr process to run out of memory when an influx of traffic arrived. The default limit is 3, but it can be configured using the `tutorial.max.forked.procs` option or the `TUTORIAL_MAX_FORKED_PROCS` environment variable.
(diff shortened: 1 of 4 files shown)


NEWS.md
```diff
@@ -12,6 +12,7 @@ learnr (development version)
 ## New features
 
 * Introduced an [experimental](https://www.tidyverse.org/lifecycle/#experimental) function `external_evaluator()` which can be used to define an exercise evaluator that runs on a remote server and is invoked via HTTP. This allows all exercise execution to be performed outside of the Shiny process hosting the learnr document.
+* For the "forked" evaluator (the default used on Linux), add a limit to the number of forked exercises that learnr will execute in parallel. Previously, this was uncapped, which could cause a learnr process to run out of memory when an influx of traffic arrived. The default limit is 3, but it can be configured using the `tutorial.max.forked.procs` option or the `TUTORIAL_MAX_FORKED_PROCS` environment variable.
```
schloerke, line 15: Do you mind linking the PRs for this news bullet and the bullet above.  (And any others we may have recently missed.) Thank you!

R/evaluators.R
```diff
@@ -34,91 +34,117 @@ inline_evaluator <- function(expr, timelimit, ...) {
 }
 
 # forked execution evaluator
-forked_evaluator <- function(expr, timelimit, ...) {
```
schloerke, line 37: Can we still define `forked_evaluator` as a result of `setup_forked_evaluator_factory(Inf)`? 

I know some people are using it externally.  Should we add a comment about that too?



... or no sympathy for people using internal methods? (Which is also a fair stance)

R/evaluators.R
```diff
@@ -128,7 +154,7 @@ forked_evaluator <- function(expr, timelimit, ...) {
 #' @import curl
 #' @export
 external_evaluator <- function(
-  endpoint = getOption("tutorial.external.host", Sys.getenv("TUTORIAL_external_evaluator_HOST", NA)),
+  endpoint = getOption("tutorial.external.host", Sys.getenv("TUTORIAL_EXTERNAL_EVALUATOR_HOST", NA)),
```
schloerke, line 157: `devtools::document()`

R/evaluators.R
```diff
@@ -34,91 +34,117 @@ inline_evaluator <- function(expr, timelimit, ...) {
...
-        closeAllConnections()
+      start = function() {
+        start_time <<- Sys.time()
+
+        doStart <- function(){
+          if (running_exercises >= max_forked_procs) {
+            # Then we can't start this job yet.
+            print("Delaying exercise execution due to forked proc limits")
+            later::later(doStart, 0.1)
+            return()
```
schloerke, line 71: This solution works really well as there are no moving parts.

I am worried as this is not a FIFO, but will randomly pick when it get's congested.  Order technically doesn't matter.  But seems odd to delay someone longer than a peer who arrives later.

Thoughts?

trestletech: That's a fair point. I guess the alternative architecture would create a queue of requests that needs to be managed? I struggled with this PR because there wasn't a great place to organize that. Until now, every exercise execution is independent and no central state needs to be coordinated. I tried to keep this implementation pretty light in that regard. But we could certainly think of a way to house more of that centralized logic somewhere to do a proper queue.

trestletech:
Good points! I think I addressed everything except the FIFO comment which I'm not sure how to fix without a more heavy-handed approach to coordination. Let me know what you think!

R/evaluators.R
schloerke: Random works for now.  Thanks!

PR #354, opened 2020-04-22
Persist cookies from the initialization call to subsequent requests

trestletech:
Adds support for persisting cookies from the initialization call to subsequent requests. This can be important if the external evaluator is behind a sticky load balancer.


tests/testthat/test-cookies.R
```diff
@@ -0,0 +1,24 @@
...
+      path = c("/", "/"),
+      secure = c(FALSE, FALSE),
+      expiration = c(1587586247L, 0L),
+      name = c("foo", "bar"),
+      value = c("123", "ftw")
+    ),
+    row.names = c(NA, -2L),
+    class = "data.frame")
+
+  f <- tempfile()
```
schloerke, line 16: on any tempfile creation, let's also `on.exit({unlink(thatfile)})`

trestletech: I like it. But we can't on.exit, we'd need to do it on session close, right? And we can only do that once the new Shiny release comes out because otherwise the reactive context isn't attached to the close handler?

Should I open an issue instead?

R/evaluators.R
```diff
@@ -246,17 +251,21 @@ internal_external_evaluator <- function(
...
-            session$userData$.external_evaluator_session_id <- sid
-            submit_req(sid)
-          }, err_callback = function(res){
-            print(res)
-            result <<- error_result("Error initiating session for external requests. Please try again later")
-          })
+          rs <- initiate(pool, paste0(endpoint, "/learnr/"),
+                         callback = function(sid, cookieFile){
+              # Stash the session ID and the cookies for future use and fire the
+              # actual request
```
schloerke, line 257: add a `onSessionEnded` callback to delete the cookie file here

tests/testthat/test-cookies.R
schloerke: Thinking of just deleting the temp files created in the tests.

tests/testthat/test-evaluators.R
```diff
@@ -287,10 +289,11 @@ test_that("", {
 
   ### Test with a bad status
   re <- internal_external_evaluator(srv$url, 5,
-    function(pool, url, callback, err_callback){ callback("badstatus") })
+    function(pool, url, callback, err_callback){ callback("badstatus", tempfile()) })
```
schloerke, line 292: Can we do the same tempfile cleanup here? (and below as well)

schloerke: Then should be good to merge. 😄

trestletech: Oops! Good catch. Fixed now

PR #355, opened 2020-04-22
Send global setup on external evaluator init

trestletech:
When initializing a new external evaluator session, send the global setup chunk so that the process can be prepared.

Also updates the openapi spec accordingly.

### DIFF
--- a/R/evaluators.R
+++ b/R/evaluators.R
@@ -280,19 +280,30 @@ internal_external_evaluator <- function(
 #' @param err_callback The callback to invoke on error. Provides one parameter:
 #'   the err'ing response
 #' @noRd
-initiate_external_session <- function(pool, url, global_setup, callback, err_callback){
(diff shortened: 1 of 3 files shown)


R/evaluators.R
```diff
@@ -279,16 +280,30 @@ internal_external_evaluator <- function(
...
 
   done_cb <- function(res){
     id <- NULL
     failed <- FALSE
     tryCatch({
       if (res$status != 200){
-        err_callback(res)
+        err_cb(res)
         return()
       }
```
schloerke, line 308: Now that `err_cb` is recursive, is it possible to move this line above the tryCatch?

It doesn't seem like this line is for the tryCatch, because if it errors, it will be called again in the error section. 

If we are suppose to catch the errors, maybe add flag to the error callback to make sure it can't be called more than once?

PR #362, opened 2020-04-30
Only initiate a single session

trestletech:
When a session is starting and restoring multiple exercises, these exercises all go through the same pipeline and independently init their own sessions. This leads to a single user spawning a dozen+ sessions, each of which then runs one exercise. 

This change creates a promise for the session identification which ensures that all of these exercises would wait on a single init() call and only then would they fire off their exercises.

This also changes the default for missing global_setup from `NULL` to `""`. The former was a violation of the openapi spec for exercises. And it makes one change to be more permissive about empty/null error messages that can pop up in some scenarios.


R/evaluators.R
```diff
@@ -251,28 +251,38 @@ internal_external_evaluator <- function(
...
+          session$userData$.external_evaluator_session_id <-
+            initiate(pool, paste0(endpoint, "/learnr/"), exercise$global_setup)
+
+          session$userData$.external_evaluator_session_id %>%
+            then(
+              onFulfilled = function(extsess){
+                # Stash the session ID and the cookies for future use and fire the
+                # actual request
+                submit_req(extsess$id, extsess$cookieFile)
+                session$onSessionEnded(function(){
```
trestletech, line 263: Move this up to 255.  And pull this promise chain out of the 253 conditional

R/setup-cache.R
```diff
@@ -17,7 +17,7 @@ get_global_setup <- function(){
     setup <- get("__setup__", envir = setup_chunks)
     return(paste0(setup, collapse="\n"))
   }
-  NULL
+  ""
```
schloerke, line 20: Will a "blank" global chunk return `"\n"`? Just wanting to make sure we can still differentiate between _empty_ and _missing_

Ex:
````md
```{r cantrememberthecorrectchunkname}
```
````

trestletech: Good point. I actually just pushed a commit that flips this. We'll use NULL for "unset" and empty quotes for empty.

Until the last commit, we had the opposite (empty string for unset and NULL for empty chunk.) It should make more sense now.

trestletech:
Just the one failing that's consistent, all others look happy.

PR #365, opened 2020-05-04
Expand external evaluator tests

trestletech:
Just updates the suite of tests we use against external evaluators

### DIFF
--- a/inst/examples/external/evaluator-tests.Rmd
+++ b/inst/examples/external/evaluator-tests.Rmd
@@ -1,5 +1,5 @@
 ---
-title: "Learnr Remote Evaluator Tests"
+title: "Learnr External Evaluator Tests"
 output: learnr::tutorial
 runtime: shiny_prerendered
 ---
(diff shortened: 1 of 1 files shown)


schloerke:
Is this a demo for other users to even see? Or just for internal testing?

Other things to possibly add:
* save a temp file
* delete the temp file
* install a package
* try to library a non-existent package
* try to run out of memory by making a large object or calling many `callr::r_bg()` processes

trestletech:
I think it's primarily for internal testing. But it would apply for anyone who wants to build an external evaluator which is why I was thinking to include it in the package.

I think the nuances of how an external evaluator would handle all those different scenarios above would just be up to their implementation. Effectively their job is to start an R session with the right context and then whatever that context is or isn't capable of is kind of out of our control. (i.e. what packages have been provisioned.) So I'm tempted to omit those.

But I do have a TODO down at the bottom for resource limitations like RAM and fork bombs. Just haven't gotten to them yet and wanted to check in what I had before I lost it.

PR #367, opened 2020-05-13
Getting simple custom question to work

bjornerstedt:
In order to get the first Custom Question Types example to work, I had to modify the code in quiz_question.Rmd. The behavior of the simple answer also becomes clearer with a second answer included.


schloerke:
Thank you!

PR #376, opened 2020-05-27
Fix question_is_correct.learnr_text comparison

### LINKED ISSUE #375: Using the `type = "learnr_text"` argument turns all answers incorrect
Using `learnr` version 0.10.1, R 3.6.3.

A minimal example for the code which recreates the bug:

````
---
title: "Tutorial"
output: learnr::tutorial
runtime: shiny_prerendered
---

```{r setup, include=FALSE}
library(learnr)
```

This works ok:
```{r works-ok}
quiz(
  question_text(
    "What is this great package you are using?",
    allow_retry = T,
    answer("learnr", correct = T)
  )
)
```

This one miss-behaves:
```{r miss-behaves}
quiz(
  question(
    "What is this great package you are using?",
    type = "learnr_text",
    allow_retry = T,
    answer("learnr", correct = T)
  )
)
```

````

Reading the documentation, I'm pretty sure I'm using the second chunk as intended, but maybe I misunderstood?

Here's how the output looks like (both should show a "Correct!" answer but the second variation fails:
bug_example

schloerke:
Fixes #375 

Also updated Actions workflow file.

`ans$label` is an `HTML()` formatted text string. This will not match.  Instead, `ans$option` should be used as it is the raw value provided by the user.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -19,18 +19,21 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - {os: windows-latest, r: '3.6'}
           - {os: windows-latest, r: 'devel'}
-          - {os: macOS-latest, r: '3.6'}
-          - {os: ubuntu-16.04, r: '3.2', cran: "https://packagemanager.rstudio.com/all/__linux__/xenial/latest"}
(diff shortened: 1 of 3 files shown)


schloerke:
Merging as is. Linked in https://github.com/rstudio/learnr/issues/383

PR #384, opened 2020-06-11
Fix #348: All exercises resolve to complete when one is completed

### LINKED ISSUE #348: All exercises resolve to complete when one is completed
If you create two exercises and put `Sys.sleep(10); print("DONE!")` inside of each, then click "Run Code" on one, wait a few seconds, and then click "Run Code" on the other, you'll see that as soon as the first one resolves, both of them appear to be done (no more spinner on "Run Code", the button and results aren't grayed out, etc.) Then the second one eventually resolves and its output will appear too.

So the behavior is correct except that the "loading" status resolves prematurely for the exercise that's the last to finish.

jcheng5:
Fixes #348.

---

Run the following, click the "Run Code" buttons one after another quickly. The progress spinner on both code chunks clear after 1 second, even though the `two` chunk doesn't complete for a few more seconds.

````
---
title: "Tutorial"
output: learnr::tutorial
runtime: shiny_prerendered
---

```{r setup, include=FALSE}
library(learnr)
```

```{r one, exercise=TRUE}
Sys.sleep(1); print("DONE!")
```

```{r two, exercise=TRUE}
Sys.sleep(5); print("DONE2!")
```
````
PR task list:
- [x] Update NEWS
- [ ] Add tests (if possible)


jcheng5:
Reviewed by @schloerke in person

Issue #356, opened 2020-04-24
Enable tutorial.error.checker to provide checks on error


garrettgman:
If a student submits code that throws an error, learnr exits the checking routine and displays the error message. This short circuit prevents the checking code from telling the student anything useful about what might have gone wrong. In short, the student is left to decipher the error message on his or her own.

This is necessary when the checking code relies on the result of the student code (if it throws an error, there is no result). But it is a missed opportunity when the checking code parses the unevaluated student code, as `gradethis::grade_code()` does. Here is an example, `grade_code()` would return `"I expected mtcars where you wrote mt"` but learnr displays the error message. (code appended at bottom)

<img width="445" alt="Screen Shot 2020-04-23 at 2 58 41 PM" src="https://user-images.githubusercontent.com/85014/80138629-121f5d00-8573-11ea-8fa1-57d583f14460.png">

To fix this, I suggest that we:
1. Add a `tutorials.error.checker` option that can be set to a grading function to run in the event that the student code throws an error. This option can be set in the same manner as the `tutorials.exercise.checker` option. It should be set to a function that can grade the student code without evaluating it, similar to `gradethis::grade_code()`.
<img width="300" alt="Screen Shot 2020-04-23 at 3 19 10 PM" src="https://user-images.githubusercontent.com/85014/80235936-5ae51d80-8628-11ea-8dff-69bd0556effe.png">

 
2. Display the result of the error checking function beneath the error message. Here is a mock up. 
<img width="445" alt="error-mockup 001" src="https://user-images.githubusercontent.com/85014/80236215-d34bde80-8628-11ea-8096-ea4bccb5a8d7.png">

Code to recreate existing example above
****
````
---
title: "Example"
output: learnr::tutorial
runtime: shiny_prerendered
---

```{r setup, include=FALSE}
library(learnr)
library(gradethis)

tutorial_options(exercise.checker = gradethis::grade_learnr)
knitr::opts_chunk$set(out.width = "100%")
```

## Welcome

### `mtcars`

To see the `mtcars` data frame, type `mtcars` in the code block below and click "Submit Answer".

```{r ex, exercise = TRUE}

```

```{r ex-solution}
mtcars
```

```{r ex-check}
grade_code("Good job!")
```
````

garrettgman:
At the moment, this is the chief reason I'm delaying wiring up an autograder to the primers.

schloerke:
Plan of action:

[...]

PR #386, opened 2020-06-12
Adds a hook for specifying an alternative to learnr:::evaluate_exercise()

dtkaplan:
## Motivation

Per [issue 356](https://github.com/rstudio/learnr/issues/356), this proposed change adds a way to replace the built-in exercise code evaluator (`learnr:::evaluate_exercise`) with one of the tutorial author's choosing. This is accomplished by adding a global option,  `learnr.alt.evaluator` that can be set to a string naming the evaluator to be used.  By default (that is, if  the option `learnr.alt.evaluator` is NULL), the built in code evaluator is used, so there is absolutely no change to the learnr code evaluation process.

Being  able to specify an alternative evaluator allows more seamless parse-time checking and provides a means to evaluate the exercise code while doing the code checking, rather than the method in learnr where the code is evaluated and the results handed off to the checker.

## Structure of change

I attempted to minimize the extent of changes to the code base by using `options()` rather than `tutorialOptions()` to set the name of the alternative evaluator. It would be perhaps more natural to the user to set the  alternative using `tutorialOptions()`, but this requires additional 
 changes to the code base. (And, since `tutorialOptions()` involves knitr hooks, I'm not confident that I would do this properly.)

I do not know enough about the new remote evaluators to determine if there is any impact on that.

## Impact of change

I am already using an alternative evaluator for a large set of tutorials I am developing. The evaluator is published as a package at `github.com/dtkaplan/learnrAlt`.  The evaluator is compatible with the facilities of the `gradethis` package. I cannot predict whether there will be many  people who want to *write* an alternative evaluator, but mine can serve as  a framework for others to work on parse-time checkers or alternative graders. @garrettgman may have some opinion about this.

## Pull Request

* Add an entry to NEWS concisely describing what you changed. DONE.

* Add unit tests in the tests/testthat directory. NOT DONE. But see the test below.

* Run Build->Check Package in the RStudio IDE, or `devtools::check()`, to make sure your change did not add any messages, warnings, or errors. DONE

## Minimal reproducible example

See the attached learnr Rmd file, which contains a trivial alternative evaluator. For a complete working evaluator, see the `github.com/dtkaplan/learnrAlt` repository.
[alt-eval.Rmd.zip](https://github.com/rstudio/learnr/files/4771823/alt-eval.Rmd.zip)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 learnr (development version)
 ===========
 
+
+
 ## Breaking Changes
 
 * Renamed the `exercise_submission` event to `exercise_result` and added the following fields:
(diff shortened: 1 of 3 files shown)


cpsievert:
Thanks for submitting this PR @dtkaplan. Instead of providing an alternative evaluator, we'll likely be provide a custom checking function to run on errors in #403

PR #387, opened 2020-06-12
temp folder for tutorial output

### LINKED ISSUE #347: Copy tutorial folder to temp folder when calling `run_tutorial`
See https://github.com/rstudio/rstudio/issues/6583

Paraphrasing...

> `learnr` attempts to render the `learnr` document directly in the installed package's folder. We need to check for write permissions and instead copy + run within the R temporary directory (since `learnr` + `learnr`-using packages might be installed into an R library that the running user doesn't have write access to)

cc @kevinushey @trestletech

bhogan-mitre:
`learnr` currently renders tutorial documents within the package directory.  This throws an error if `learnr` is installed in a directory that the user does not have write permission in.

This PR renders documents in a temp directory instead.

Closes #347 

PR task list:
- [ x ] Update NEWS
- [ ] Add tests (if possible)

### DIFF
--- a/R/run.R
+++ b/R/run.R
@@ -51,13 +51,17 @@ run_tutorial <- function(name = NULL, package = NULL, shiny_args = NULL) {
     )
   }
 
+  # set rmarkdown args to render in tmp dir
+  render_args <- list(output_dir = tempdir())
+
   # run within tutorial wd
   withr::with_dir(tutorial_path, {
     if (!identical(Sys.getenv("SHINY_PORT", ""), "")) {
       # is currently running in a server, do not allow for prerender (rmarkdown::render)
       withr::local_envvar(c(RMARKDOWN_RUN_PRERENDER = "0"))
     }
-    rmarkdown::run(file = NULL, dir = tutorial_path, shiny_args = shiny_args)
+    rmarkdown::run(file = NULL, dir = tutorial_path, shiny_args = shiny_args,
+                   render_args = render_args)
   })
 }
 
--- a/R/run.R
+++ b/R/run.R
@@ -52,7 +52,12 @@ run_tutorial <- function(name = NULL, package = NULL, shiny_args = NULL) {
   }
 
   # set rmarkdown args to render in tmp dir
-  render_args <- list(output_dir = tempdir())
+  temp_output_dir <- tempdir()
+  render_args <- list(
+    output_dir = temp_output_dir,
+    intermediates_dir = temp_output_dir,
+    knit_root_dir = temp_output_dir
+  )
 
   # run within tutorial wd
   withr::with_dir(tutorial_path, {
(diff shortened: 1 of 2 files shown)


schloerke:
@bhogan-mitre Do you mind signing the CLA? Once you've done that, the check will turn green.  Thank you!

R/run.R
```diff
@@ -51,13 +51,22 @@ run_tutorial <- function(name = NULL, package = NULL, shiny_args = NULL) {
     )
   }
 
+  # set rmarkdown args to render in tmp dir
+  temp_output_dir <- tempdir()
```
schloerke, line 55: Let's namespace the folders with the package and name inside a `"learnr"` folder.  (And create the folder if it doesn't exist.

```suggestion
  temp_output_dir <- file.path(tempdir(), "learnr", package, name)
  if (!dir.exists(temp_output_dir)) {
    dir.create(temp_output_dir, recursive = TRUE)
  }
```

R/run.R
```diff
@@ -51,13 +51,22 @@ run_tutorial <- function(name = NULL, package = NULL, shiny_args = NULL) {
     )
   }
 
+  # set rmarkdown args to render in tmp dir
```
schloerke, line 54:
```suggestion
  # Set rmarkdown args to render in tmp dir
  # This will cause the tutorial to be re-rendered in each R session
```

R/run.R
```diff
@@ -51,13 +51,22 @@ run_tutorial <- function(name = NULL, package = NULL, shiny_args = NULL) {
...
+  temp_output_dir <- tempdir()
+  render_args <- list(
+    output_dir = temp_output_dir,
+    intermediates_dir = temp_output_dir,
+    knit_root_dir = temp_output_dir
+  )
```
schloerke, line 60: Trying to be on the safe side... rather than changing the default behavior and forcing users to always recompile on a new R session, let's attempt to write to the local directory.  If we can not, then we will use the tmp directory.

```suggestion
  render_args <-
    tryCatch({
      local({
        # try to save a file to check for write permissions
        tmp_save_file <- file.path(tutorial_path, "__leanr_test_file")
        # make sure it's deleted
        on.exit({
          if (file.exists(tmp_save_file)) {
            unlink(tmp_save_file)
          }
        }, add = TRUE)
        # write to the test file
        cat("test", file = tmp_save_file)
        # if no errors have occurred, return an empty list of render_args
        list()
      })
    }, error = function(e) {
      # Could not write in the tutorial folder

      # Set rmarkdown args to render in tmp dir
      # This will cause the tutorial to be re-rendered in each R session
      temp_output_dir <- file.path(tempdir(), "learnr", package, name)
      if (!dir.exists(temp_output_dir)) {
        dir.create(temp_output_dir, recursive = TRUE)
      }
      list(
        output_dir = temp_output_dir,
        intermediates_dir = temp_output_dir,
        knit_root_dir = temp_output_dir
      )
    })
```

schloerke:
Closing in favor of #412

PR #398, opened 2020-07-14
Add event handler system

### LINKED ISSUE #357: Trigger actions when student advances between sections
I'm settling more and more on tutorials that use progressive reveal (`progressive: true` in the YAML) displayed within or alongside the IDE. When combined with functionality with the `rstudioapi` package, this provides a convenient way to teach students not just how to write code, but also how to interact with the IDE.

**For these tutorials, it would be incredibly useful to trigger code whenever a student advances from one subsection to the next.** i.e. it would be incredibly useful to provide code that is run when a subsection begins (`pre`) and/or to provide code that is run when a section ends (`post`).

Here is an example use case:

1. The student clicks the "Continue" button to move to the next subsection, which will introduce the `mpg` dataset. The subsections that follow will also discuss the `mpg` dataset. I would like to display the dataset in a way that persists as the student moves from subsection to subsection. I would also like to familiarize the student with the RStudio IDE data viewer.

2. I have provided the following code for learnr to run when the tutorial advances to the new subsection, `rstudioapi::sendToConsole("View(ggplot2::mpg)", echo = FALSE, focus = FALSE)`.  I've provided it as an R code chunk with a special label, or a special option, or I've provided it in `{` after the subsection title, or by some other documented method

3. learnr runs the code as it advances to the new subsection, which opens `mpg` in the data viewer.

wch:
This adds a general-purpose event handler system. Closes #357.

It adds:

* `event_register_handler()`
* `one_time()`

It also converts many of the existing event-like functions into to true event handlers.

It adds a new event, `"section_viewed"`, which is triggered every time a section becomes visible.

The `event_register_handler()` function can be used to execute code every time an event occurs. For example, this will print a message to the console every time any section becomes visible.

	```{r context="server-start"}
	event_register_handler("section_viewed", function(session, event, data) {
	  message("event: section_viewed: ", data$sectionId)
	})
	```

The `one_time()` function can be used to only execute code the first time an event happens AND some condition is satisfied. This can be used to execute code the only the first time a section is viewed:

	```{r context="server-start"}
	event_register_handler("section_viewed",
	  function(session, event, data) {
	    one_time(
	      session,
	      cond = data$sectionId == "section-exercise-with-hint",
	      expr = { message("Seeing ", data$sectionId, " for the first time.") }
	    )
	  }
	)
	```

*****

### Usage notes

If a section doesn't have a name, then one will be auto-generated. In the example below, the first one will trigger a `section_viewed` event where `data$sectionId` is `"section-section"` and the second will  be `"section-section-1"`.

```
###

foo

###

bar

```

You can also add specific section IDs without having any text show up. This is better in practice, so that adding a section in the middle doesn't alter up later section IDs that you're looking for. In the example below, the first one will be `"section-first"` and the second will be `"section-second"`.

```
### {#first}

foo

### A title here {#second}

bar

```

*****

### Implementation notes

The events starts in tutorial-format.js. There are `shown` and `hidden` events for the topics and sections. Unfortunately, they fire too often, and sometimes (for example) a `shown` event occurs when a section isn't actually visible. The code in that file converts those trigger-happy events into _state_, in the form a Shiny input value. It sets ``input$`tutorial-visible-sections` `` to a character vector that contains the IDs of the visible sections.

On the R side, the states from ``input$`tutorial-visible-sections` `` are converted back to events, using an observer which calls `event_trigger("section_viewed", ....)`.


R/event.R
```diff
@@ -0,0 +1,92 @@
...
+  # Handlers for this named event
+  handlers <- event_handlers[[event]]
+
+  # Invoke all the callbacks for this event.
+
+  # NOTE: These are not wrapped in try-catch, so an error will stop all the rest
+  # of the callbacks from executing.
+  for (handler in handlers) {
+    handler(session, event, data)
+  }
```
schloerke, line 91: Would like to add a tryCatch around this to make sure another package doesn't prevent user additions.

```suggestion
  for (handler in handlers) {
    tryCatch(
      {
        handler(session, event, data)
      }, 
      error = function(e) {
        traceback() # some form of traceback on the error?
      }
    )
  }
```

R/event.R
```diff
@@ -0,0 +1,92 @@
...
+  event_handlers[[event]][[id]] <- NULL
+  invisible(TRUE)
+}
+
+
+event_trigger <- function(session, event, data = list()) {
+  if (is.null(event_handlers[[event]])) {
+    return(invisible())
+  }
+
```
wch, line 81: Call `record_event()` here.

R/events-record.R
```diff
@@ -0,0 +1,167 @@
...
+      session  = session,
+      label    = data$label,
+      question = data$question,
+      answer   = data$answer
+    )
+  }
+)
+
+event_register_handler(
+  "reset_question_submission",
```
wch, line 51: In NEWS, mention that this new event type is a breaking change.

R/events-record.R
```diff
@@ -0,0 +1,167 @@
...
+      event = "section_skipped",
+      data = data
+    )
+
+    # save for later replay
+    save_section_skipped(session = session, sectionId = data$sectionId)
+  }
+)
+
+event_register_handler(
```
wch, line 91: Can remove this handler (after `record_event` is called from `event_trigger`)

R/events-record.R
```diff
@@ -0,0 +1,167 @@
...
+  }
+)
+
+
+event_register_handler(
+  "exercise_result",
+  function(session, event, data) {
+
+    # Set some defaults if needed.
+    if (! "checked" %in% names(data))
```
wch, line 105: Can remove these defaults - no longer necessary.

R/events-record.R
```diff
@@ -0,0 +1,167 @@
...
+  ))
+}
+
+broadcast_question_event_to_client <- function(session, label, answer) {
+  broadcast_progress_event_to_client(session = session,
+                                     event = "question_submission",
+                                     data = list(label = label, answer = answer))
+}
+
+event_register_handler(
```
wch, line 27: Create a function called `register_default_handlers` which does all these calls to `event_register_handler`. In `tutorial()`, call `event_handlers_reset()` (or something like that) and then `register_default_handlers()`.

schloerke: Thinking about this more, `event_handlers_reset()` and `register_default_handlers()` should be called inside `initialize_tutorial()`, not in `tutorial()`. This should be done as a shiny pre-rendered chunk.

The setup chunk cache is already being cleared here:

https://github.com/rstudio/learnr/blob/659b1bbe82278fd027c745d1af41a52962c2f716/R/initialize.R#L36-L41

R/event.R
wch: I've added a `tryCatch`. Not sure yet how to get a traceback, but I'm looking into it.

R/events-record.R
wch: This is now done at package load time.

inst/rmarkdown/templates/tutorial/resources/tutorial-format.js
```diff
@@ -408,6 +442,43 @@ $(document).ready(function() {
     }
   }
 
+  // Keep track of which sections are currently visible. When this changes
+  var visibleSections = [];
```
wch, line 446: Rename variable to be consistent with function name

inst/rmarkdown/templates/tutorial/resources/tutorial-format.js
```diff
@@ -408,6 +442,43 @@ $(document).ready(function() {
...
+      }
+    }
+  }
+
+  function sendVisibleSections() {
+    // This function may be called several times in a tick, which results in
+    // many calls to Shiny.setInputValue(). That shouldn't be a problem since
+    // those calls are deduped; only the last value gets sent to the server.
+    if (Shiny && Shiny.setInputValue) {
+      Shiny.setInputValue("tutorial-visible-sections", visibleSections);
```
wch, line 473: Maybe store topic ID along with section ID. Would also need to change `section_skipped`.

inst/rmarkdown/templates/tutorial/resources/tutorial-format.js
```diff
@@ -453,6 +524,7 @@ $(document).ready(function() {
 
     // handle progress events
     tutorial.onProgress(function(progressEvent) {
+      console.log('progress: ' + progressEvent.event);
```
wch, line 527: Remove this.

R/events_default.R
```diff
@@ -0,0 +1,176 @@
...
+broadcast_question_event_to_client <- function(session, label, answer) {
+  broadcast_progress_event_to_client(session = session,
+                                     event = "question_submission",
+                                     data = list(label = label, answer = answer))
+}
+
+
+register_default_event_handlers <- function() {
+
+  event_register_handler("session_start",
```
wch, line 18: Put `session_start` on next line.

R/tutorial-format.R
```diff
@@ -127,6 +127,12 @@ tutorial <- function(fig_width = 6.5,
   # set 1000 as the default maximum number of rows in paged tables
   knitr_options$opts_chunk$max.print <- 1000
 
+  # Clear all event handlers and register the default event handlers.
+  # TODO: Currently, the `_once` handler isn't restored if user starts over.
+  #       Make this work somehow.
+  event_handlers_reset()
```
wch, line 133: Remove these lines

R/events_default.R
```diff
@@ -0,0 +1,176 @@
...
+      # notify client side listeners
+      broadcast_progress_event_to_client(session, "video_progress", data)
+
+      # save for later replay
+      save_video_progress(session, data$video_url, data$time, data$total_time)
+    }
+  )
+}
+
+register_default_event_handlers()
```
wch, line 175: Move into `.onLoad()` or `tutorial()`

inst/rmarkdown/templates/tutorial/resources/tutorial-format.js
wch: I've decided against making this change.

inst/rmarkdown/templates/tutorial/resources/tutorial-format.js
wch: Not doing topic ID, since we found that section IDs are always unique.

R/events.R
```diff
@@ -1,199 +1,188 @@
...
+#' }
+#' @export
+one_time <- function(session, cond, expr, label = deparse(substitute(cond))) {
+  # This is meant to be called within an event handler, instead of being
+  # implemented as an event handler that removes itself. The reason that the
+  # latter method isn't used is because it's possible that two simultaneous user
+  # sessions will use be running, and these event handlers are installed at an
+  # app-level (not session-level) scope.
+
+    if (!isTRUE(cond)) {
```
cpsievert, line 165:
```suggestion
  if (!isTRUE(cond)) {
```

PR #403, opened 2020-07-21
Checking for exercise errors

### LINKED ISSUE #356: Enable tutorial.error.checker to provide checks on error

cpsievert:
Closes #356. Note that this allows checking of (evaluation) errors without any additional options. Here's a minimal example:

````
---
title: "Checking errors"
output: learnr::tutorial
runtime: shiny_prerendered
---

```{r setup, include=FALSE}
library(learnr)	
library(gradethis)
# remotes::install_github("rstudio-education/gradethis")
tutorial_options(
  exercise.checker = grade_learnr, 
  exercise.error.checker = grade_learnr_error
)
```

* Throw an error message of `"boom"`:

```{r grade_error, exercise = TRUE}

```

```{r grade_error-solution}
stop("boom")
```

```{r grade_error-code-check}
grade_code("Nice error!")
```
````

### Known breaking change

Now, if a `-code-check` chunk exists, then we no longer display the results of the exercise (press "Run Code" if you want results and/or do the grading in the `-check` chunk)

### TODO

- [x] Update news (and mention breaking change)

### CODE REVIEW

File: R/options.R [hunk 1]
```diff
@@ -22,9 +24,12 @@ tutorial_options <- function(exercise.cap = "Code",
                              exercise.timelimit = 30,
                              exercise.lines = NULL,
                              exercise.checker = NULL,
+                             exercise.error.checker = NULL,
+                             exercise.parse.checker = NULL,
```
cpsievert, line 28: This option really wants to live in gradethis, but it doesn't seem worth re-exporting

cpsievert: Remove the changes here in `R/options.R`

R/exercise.R
```diff
@@ -176,215 +176,144 @@ evaluate_exercise <- function(exercise, envir, evaluate_global_setup = FALSE) {
...
-  err <- NULL
-  get_checker <- function() {
-    checker <- exercise$options$exercise.checker
-    if (is.function(checker)) {
-      environment(checker) <- envir_prep
-    } else if (!is.null(checker)) {
-      warning("Found a exercise.checker that isn't a function", call. = FALSE)
-      checker <- NULL
+  # Setup a temporary directory for rendering the exercise
+  exercise_dir <- tempfile(pattern = "learnr-tutorial-exercise")
```
cpsievert, line 191: Because Windows

```suggestion
  exercise_dir <- tempfile(pattern = "lnr-ex")
```

R/options.R
```diff
@@ -22,6 +22,7 @@ tutorial_options <- function(exercise.cap = "Code",
                              exercise.timelimit = 30,
                              exercise.lines = NULL,
                              exercise.checker = NULL,
+                             exercise.error.checker = NULL,
```
schloerke, line 25: Needs param line above

PR #413, opened 2020-07-31
Disable completion of R code inside quotes, closes #401

### LINKED ISSUE #401: Ace's file autocomplete lists files in wrong directory
If I run a tutorial from learnr's src directory, I see

Screen Shot 2020-07-17 at 10 04 48 AM

This autocomplete feature should either be disabled or set to the _exercise's_ working directory

cpsievert:
See #401 for motivation.

This disables completion of R code inside quotes (when `exercise.completion = TRUE`), by setting `quotes = FALSE` in `rc.settings()`...the docs for `quotes` says:

> Enables completion in R code when inside quotes. This normally leads to filename completion, but can be otherwise depending on context (for example, when the open quote is preceded by ?), help completion is invoked. Setting this to FALSE relegates completion to the underlying completion front-end, which may do its own processing (for example, readline on Unix-alikes will do filename completion).

Thus, this might be a slightly over-agressive way to prevent filenames from being displayed. Note that there is also a `files` argument that we could set to `FALSE`, but for some reason that arg is deprecated (maybe it's effectively the same as `quotes`?)

### DIFF
--- a/R/http-handlers.R
+++ b/R/http-handlers.R
@@ -187,7 +187,7 @@ register_http_handlers <- function(session, metadata) {
     settings <- utils::rc.settings()
     utils::rc.settings(ops = TRUE, ns = TRUE, args = TRUE, func = FALSE,
                        ipck = TRUE, S3 = TRUE, data = TRUE, help = TRUE,
-                       argdb = TRUE, fuzzy = FALSE, files = TRUE, quotes = TRUE)
+                       argdb = TRUE, fuzzy = FALSE, files = TRUE, quotes = FALSE)
     on.exit(do.call(utils::rc.settings, as.list(settings)), add = TRUE)
 
     # temporarily attach global setup to search path
--- a/R/http-handlers.R
+++ b/R/http-handlers.R
@@ -187,7 +187,7 @@ register_http_handlers <- function(session, metadata) {
     settings <- utils::rc.settings()
     utils::rc.settings(ops = TRUE, ns = TRUE, args = TRUE, func = FALSE,
                        ipck = TRUE, S3 = TRUE, data = TRUE, help = TRUE,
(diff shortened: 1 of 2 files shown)


R/http-handlers.R
```diff
@@ -187,7 +187,7 @@ register_http_handlers <- function(session, metadata) {
     settings <- utils::rc.settings()
     utils::rc.settings(ops = TRUE, ns = TRUE, args = TRUE, func = FALSE,
                        ipck = TRUE, S3 = TRUE, data = TRUE, help = TRUE,
-                       argdb = TRUE, fuzzy = FALSE, files = TRUE, quotes = TRUE)
+                       argdb = TRUE, fuzzy = FALSE, files = TRUE, quotes = FALSE)
```
schloerke, line 190: Let's also set `files` to `FALSE`, even though it's deprecated.

```suggestion
                       argdb = TRUE, fuzzy = FALSE, files = FALSE, quotes = FALSE)
```

R/http-handlers.R
```diff
@@ -187,7 +187,7 @@ register_http_handlers <- function(session, metadata) {
     settings <- utils::rc.settings()
     utils::rc.settings(ops = TRUE, ns = TRUE, args = TRUE, func = FALSE,
                        ipck = TRUE, S3 = TRUE, data = TRUE, help = TRUE,
-                       argdb = TRUE, fuzzy = FALSE, files = TRUE, quotes = TRUE)
+                       argdb = TRUE, fuzzy = FALSE, files = FALSE, quotes = FALSE)
```
cpsievert, line 190: Add a comment of: ``# If and when exercises have access to files, then set `files = TRUE, quotes = TRUE`, sanitize, and evaluate in the exercise dir``

PR #420, opened 2020-08-13
Fixes for learnr to external evaluator

nischalshrestha:
### Problems

When using the  "new" (commit >= bd29e0c145e36211c3f86987209c767c731a1abc) learnr against an "old" (commit: cf5fe34eaee039986596e1578bac0d5ffdb01d06) learnr _on the external evaluator_,  there are evaluator and feedback return   issues:

- the evaluator always returned four empty lines for every exercise. 
- the engine for R exercise needs to be capital "R" not "r"
- the `exercise$options$code` for some reason gets reused as exercise code instead of `exercise$code`, which causes the external server to always run the original exercise code (from knitr chunk) instead of the user code. 
- regular setup code will not work because the `exercise$setup` is always `NULL` since we now store all setup chunks in a `exercise$chunks` list.
- `exercise.checker` for new learnr caches function which causes issues on deserializing on external evaluator.

This is because the exercise structure for new learnr changed by: 1) forwarding all knitr options to evaluation, which causes a problem when sending some `exercise$options` fields like `eval` (produces for empty lines), and 2) `exercise.checker` was treated as a function when caching exercise.

### Fixes

- In `knitr-hooks.R`, removing the `exercise$eval` field fixes the empty lines issue, while removing `exercise$options$code` makes sure the old learnr executes user code, **NOT** original knitr chunk code.
- The `exercise$options$engine` will be removed on external evaluator (as both old learnr and new learnr do not directly use it).
- To fix regular setup* code for exercises, I added back an `exercise$setup` in `knitr-hooks.R` (another field for cache), which contains either `NULL` if there is not setup code, or a `character` containing all of the setup code required (newline separated). 
- `exercise.checker` for new learnr is back to being a character, so now all learnr versions will stick to this format.

*Note: this means that regular exercise setup will work, but chained will not, especially if there is non-R setup chunks.

You can test the solution by running `evaluator-tests.Rmd` learnr document and going through all tests in the tutorial.


R/utils.R
```diff
@@ -95,7 +95,7 @@ is_tags <- function(x) {
 }
 
 knitr_engine <- function(engine) {
-  tolower(engine %||% "r")
```
trestletech, line 98: It seems desirable to me to standardize on upper/lower. What's the motivation for this? 

Should external evaluators just be resilient to upper/lower case here?

R/knitr-hooks.R
```diff
@@ -277,10 +277,28 @@ install_knitr_hooks <- function() {
         check_chunk <- get_knitr_chunk(paste0(options$label, "-check"))
         solution <- get_knitr_chunk(paste0(options$label, "-solution"))
 
-        # remove class of "knitr_strict_list" so serializing works properly for external evaluators
+        # remove specific option fields and construct an exercise$setup to make old learnr on external evaluator server happy
+        # sending an eval field for options causes the 4 empty lines issue (https://github.com/rstudio/rmote/issues/36)
+        options$eval <- NULL
+        # `code` needs to be cleared since the original code of exercise seems to get used every time we evaluate exercise
+        # instead of using user code via `exercise$code`
+        options$code <- NULL
```
trestletech, line 285: Squashing these two options also feels like something we could consider doing in the external evaluator.

R/knitr-hooks.R
```diff
@@ -277,10 +277,28 @@ install_knitr_hooks <- function() {
...
+        # remove class of "knitr_strict_list" so (de)serializing works properly for external evaluators
         class(options) <- NULL
-
-        exercise_cache <- list(chunks = all_chunks,
+        # we collect all the setup code to make exercise compatible with old learnr
+        # note: this means that chained setup chunks will not work for non-R exercises
+        all_setup_code =
+          if (length(all_chunks) > 1) {
```
trestletech, line 291: I haven't seen this assignment via if/else construct before in R. Are we using that elsewhere in this repo?

If not, I think something like the following might be more familiar to R folks:

```r
all_setup_code <- NULL
if (length(all_chunks) > 1) {
  ...
  all_setup_code <- whatever
}
```

R/utils.R
nischalshrestha: Old learnr wanted a capital "R" for some reason when trying to execute exercise. You could have external evaluators handle this but my line of reasoning was, this function is used as part of setting up the structure initially, and doesn't break new learnr.

R/knitr-hooks.R
nischalshrestha: I would be open to moving this squashing logic in the external evaluator as it does seem odd to see from the learnr perspective.

R/utils.R
trestletech: I can get on board with that. I guess my argument would be that we should change `tolower` to `toupper` so that we're still standardized on a single case.

trestletech:
We discussed and the root of the problem here is that learnr recently attached arbitrary properties it gets from knitr to each exercise. It turns out that some of the properties cause problems when interacting with old learnrs in a remote evaluator. 

Given that these changes were recently introduced and we're basically just scrubbing a couple of known-problematic fields off of the exercise, this doesn't feel so out of place to me.

I vote we move forward with this approach but moving forward do some work to think about how we can better capture what structure an exercise is supposed to have so that others (e.g. remote evaluators) can confidently interact with them.

R/utils.R
nischalshrestha: Definitely. I'll try capital case and see if anything breaks, and if not, let's do `toupper`

nischalshrestha: Going back to standardize on lowercase engine, we will simply remove `exercise$options$engine` as both old and new learnr does not make use of it for evaluation purposes.

R/knitr-hooks.R
nischalshrestha: This is now squashed in the external evaluator not in learnr.

R/knitr-hooks.R
```diff
@@ -271,16 +271,27 @@ install_knitr_hooks <- function() {
...
 
         code_check_chunk <- get_knitr_chunk(paste0(options$label, "-code-check"))
         check_chunk <- get_knitr_chunk(paste0(options$label, "-check"))
         solution <- get_knitr_chunk(paste0(options$label, "-solution"))
 
-        # remove class of "knitr_strict_list" so serializing works properly for external evaluators
+        # remove class of "knitr_strict_list" so (de)serializing works properly for external evaluators
         class(options) <- NULL
+        # we collect all the setup code to make exercise compatible with old learnr
+        # note: this means that chained setup chunks will not work for non-R exercises
```
nischalshrestha, line 284:
```suggestion
        # note: this means that chained setup chunks will not work for non-R exercises
        # Remove this after v0.13 is released
```

trestletech (approved):

PR #426, opened 2020-08-28
Remove exercise.error.checker in favor of exercise.error.check.code and -error-check chunks

### LINKED ISSUE #356: Enable tutorial.error.checker to provide checks on error

### LINKED ISSUE #425: exercise.error.checker should pass through error messages on Run Code
When the tutorial option `exercise.error.checker` is set, learnr returns an unintelligible error when a student runs bad code. Here is an example. This happens when the student clicks Run Code.

Screen Shot 2020-08-27 at 9 44 16 AM

When this option is removed, learnr displays the correct, intelligible error message when the student clicks Run Code. This is the error message you would get if you run this code at the command line.

Screen Shot 2020-08-27 at 9 44 38 AM

This may be a gradethis bug. It is clear that `gradethis::grade_learnr_error` is related to the behavior, but it is unclear if `gradethis::grade_learnr_error` should be invoked when the student clicks Run Code. See https://github.com/rstudio-education/gradethis/issues/140.

Demo .Rmd:

````
---
output: learnr::tutorial
runtime: shiny_prerendered
---

```{r setup, include=FALSE}
library(learnr)
library(gradethis)
library(tidyverse)

tutorial_options(
  exercise.checker = gradethis::grade_learnr,
  exercise.error.checker = gradethis::grade_learnr_error
)
```

## Aesthetics

### 

Map the chape aesthetic to a continuous variable. What happens?

```{r cty3, exercise = TRUE, exercise.eval = FALSE}
# Map cty to shape
ggplot(data = mpg, mapping = aes(x = displ, y = hwy)) + 
  geom_point()
```

```{r cty3-solution}
# Map cty to shape
ggplot(data = mpg, mapping = aes(x = displ, y = hwy, shape = cty)) + 
  geom_point()
```

```{r cty3-code-check}
grade_code("ggplot2 will map shapes to discrete variables, but not continuous variables; but that makes sense: shapes are discrete. Click Run Code to see the error message that you'll get when you map shapes to a continuous variable.")
```

````

### PR DESCRIPTION
**Pull request by cpsievert**
Fixes #356 
Fixes #425 
Fixes https://github.com/rstudio-education/gradethis/issues/140
Related https://github.com/rstudio-education/gradethis/pull/143

After feedback from @garrettgman, and discussion with @schloerke, it's becoming apparent the `exercise.error.checker` probably isn't the right approach to exercise (evaluation) error checking.

This PR replaces `exercise.error.checker` with a slightly different option, `exercise.error.check.code`, which defines what checking code to run when an exercise error occurs. This option, in some sense, provides a global default value, which can be overridden for a particular exercise with a `-error-check` chunk. Overall this seems like a much cleaner solution than `exercise.error.checker`  since it decouples error checking code from `-code-check` (pre-evaluation) and `-check` (post evaluation)

````
```{r setup, include=FALSE}
library(learnr)
library(gradethis)

tutorial_options(
  exercise.checker = gradethis::grade_learnr,
  exercise.error.check.code = "gradethis::grade_code()"
)
```

* Enter `mtcars`

```{r mtcars, exercise = TRUE}

```

```{r mtcars-solution}
mtcars
```

```{r mtcars-error-check}
grade_code(incorrect = "This code produces an error (press 'Run Code' to see it).", glue_incorrect = "{ .message } { .incorrect }")
```

```{r mtcars-check}
grade_result(
  fail_if(~identical(.result, cars), "This is the cars (not mtcars) dataset."),
  pass_if(~identical(.result, mtcars))
)
```
````

### DIFF
--- a/R/exercise.R
+++ b/R/exercise.R
@@ -65,6 +65,7 @@ setup_exercise_handler <- function(exercise_rx, session) {
     # - solution
     # - engine
     exercise <- append(exercise, get_exercise_cache(exercise$label))
+    # If there is no locally defined error check code, look for globally defined error check option
     exercise$error_check <- exercise$error_check %||% exercise$options$exercise.error.check.code
     if (!isTRUE(exercise$should_check)) {
       exercise$check <- NULL
--- a/R/exercise.R
+++ b/R/exercise.R
@@ -61,13 +61,15 @@ setup_exercise_handler <- function(exercise_rx, session) {
     exercise$global_setup <- get_global_setup()
     # retrieve exercise cache information:
     # - chunks (setup + exercise) for the exercise to be processed in `evaluate_exercise`
-    # - checker code (check, code-check)
+    # - checker code (check, code-check, error-check)
     # - solution
     # - engine
     exercise <- append(exercise, get_exercise_cache(exercise$label))
+    exercise$error_check <- exercise$error_check %||% exercise$options$exercise.error.check.code
     if (!isTRUE(exercise$should_check)) {
       exercise$check <- NULL
       exercise$code_check <- NULL
+      exercise$error_check <- NULL
     }
     # variable has now served its purpose so remove it
     exercise$should_check <- NULL
@@ -391,12 +393,11 @@ render_exercise <- function(exercise, envir, envir_prep) {
     if (grepl(pattern, msg, fixed = TRUE)) {
       return(exercise_result_timeout())
     }
-    error_check_code <- exercise$error_check %||% exercise$options$exercise.error.check.code
-    if (length(error_check_code)) {
+    if (length(exercise$error_check)) {
       # Run the condition through an error checker (the exercise could be to throw an error!)
       checker_feedback <- try_checker(
         exercise, "exercise.checker",
-        check_code = error_check_code,
+        check_code = exercise$error_check,
         envir_result = envir,
         evaluate_result = evaluate_result,
         envir_prep = envir_prep,
--- a/R/exercise.R
+++ b/R/exercise.R
@@ -391,20 +391,22 @@ render_exercise <- function(exercise, envir, envir_prep) {
     if (grepl(pattern, msg, fixed = TRUE)) {
       return(exercise_result_timeout())
     }
-    # Run the condition through an error checker (the exercise could be to throw an error!)
-    checker_feedback <- try_checker(
-      exercise, "exercise.error.checker",
-      check_code = exercise$check,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: R/exercise.R [hunk 1]
```diff
@@ -61,13 +61,15 @@ setup_exercise_handler <- function(exercise_rx, session) {
     exercise$global_setup <- get_global_setup()
     # retrieve exercise cache information:
     # - chunks (setup + exercise) for the exercise to be processed in `evaluate_exercise`
-    # - checker code (check, code-check)
+    # - checker code (check, code-check, error-check)
     # - solution
     # - engine
     exercise <- append(exercise, get_exercise_cache(exercise$label))
+    exercise$error_check <- exercise$error_check %||% exercise$options$exercise.error.check.code
```
schloerke, line 68:
```suggestion
    # If there is no locally defined error check code, look for globally defined error check option
    exercise$error_check <- exercise$error_check %||% exercise$options$exercise.error.check.code
```

Issue #382, opened 2020-06-06
Question requiring at leat one correct (instead of all correct)


blaiseli:
I haven't seen this request, but maybe I didn't look carefully enough: I tried to make a `question` where several answers are considered correct, and the behaviour I obtained was that the only correct solution was to check all correct answers. My intention was that it would be considered correct if at least one of the correct answers was checked (and none of the wrong answers).

For instance:
```
question("Select at least one TRUE expression",
         answer("`2 < 4`", correct = TRUE, message = "`!FALSE` is also correct"),
         answer("`!FALSE`", correct = TRUE, message = "`2 < 4` is also correct"),
         answer("`TRUE & (2 < 1)`"),
         allow_retry = TRUE
         )
```

Selecting only one of the correct answers is considered wrong. What would be nice is an `any_correct` option that would result in the desired behaviour.

A related feature would be to allow multiple choice square check boxes even when only one answer is correct, in order to avoid giving the hint that only one answer is correct.

PR #434, opened 2020-09-25
Add new anybox question type to solve #382

elmstedt:
Solves #382

This creates a question type which has multiple correct questions but for which it is not required to select all of the correct options before proceeding. A successfully answered question will add messages indicating which correct responses were missed as well as which incorrect responses were selected.

In order to pass the additional arguments `min_right` and `max_wrong` to `question_anybox()`, I added them as attributes to the question object returned by `learnr::question()`. This feels a bit kludgy, but I wanted to not alter the structure of the question object to avoid any potential conflicts. Giving the object two additional attributes seemed the least invasive. 

```` r
---
title: "Tutorial"
output: learnr::tutorial
runtime: shiny_prerendered
---

```{r setup, include=FALSE}
library(learnr)
```

## Topic 1

```{r two_right_one_wrong, echo = FALSE}
question_anybox(
  "Select at least two toppings that belong on a Margherita Pizza (and no more than one that doesn't):",
  answer("tomato", correct = TRUE, message = "Tomatoes too!"),
  answer("mozzarella", correct = TRUE, message = "Don't forget the cheese!"),
  answer("basil", correct = TRUE, message = "Basil gives it a distinctive flavor!"),
  answer("extra virgin olive oil", correct = TRUE, message = "You need olive oil too!"),
  answer("pepperoni", message = "Pepperoni is a great topping! ... just not on a Margherita Pizza"),
  answer("onions", message = "Onions!? No and yuck!"),
  answer("bacon", message = "Bacon doesn't belong here!"),
  answer("spinach", message = "Spinach? With Olive Oil? Only if you're Popeye!"),
  random_answer_order = TRUE,
  allow_retry = TRUE,
  try_again = "Be sure to select all four toppings!",
  min_right = 2,
  max_wrong = 1
)
```
````
PR task list:
- [X] Update NEWS
- [X] Add tests (if possible)
- [X] Update documentation with `devtools::document()`

### CODE REVIEW

File: R/question_anybox.R [hunk 1]
```diff
@@ -0,0 +1,131 @@
...
+  min_right <- max(attr(question, "min_right"), 1)
+  max_wrong <- max(attr(question, "max_wrong"), 0)
+  ans <- question[["answers"]]
+  anss <- vapply(ans, `[[`, character(1), "option")
+  corr <- vapply(ans, `[[`, logical(1), "correct")
+  cor_ans <- anss[corr]
+  check <- match(value, cor_ans)
+  right <- cor_ans[stats::na.omit(check)]
+  wrong <- ans[match(setdiff(value, cor_ans), anss)]
+  missed <- ans[match(setdiff(cor_ans, value), anss)]
```
schloerke, line 94: Cosmetic request: Could you expand on the variable names being created? Thank you

schloerke (changes requested):

schloerke:
I also don't know if this logic should be merged into the default checkbox logic, or if it should be made into a separate exported function.

PR #429, opened 2020-09-04
Display engine images in the exercise caption

schloerke:
PR task list:
- [x] Update NEWS
- [NA] Add tests (if possible)
- [x] Update documentation with `devtools::document()`

```r
remotes::install_github("rstudio/learnr#429")
rmarkdown::run("learnr_429.Rmd")
```

<details>
  <summary> learnr_429.Rmd </summary>

````
---
title: "Hello, Tutorial!"
author: "J.J. Allaire"
date: "March 1st, 2017"
output: learnr::tutorial
runtime: shiny_prerendered
# Do not index/display tutorial by setting `private: true`
# private: true
description: >
  This is a demo tutorial.
---

```{r setup, include=FALSE}
library(learnr)
library(gradethis)
gradethis::gradethis_setup()
```

The following code computes the answer to 1+1. Change it so it computes 2 + 2:

```{r addition, exercise=TRUE}
1 + 1
```




```{r test, exercise=TRUE}
2+2
```

```{r test-solution}
#solution
2 + 2
```

```{r test-hint-1}
#hint1
```


```{r test-hint-2}
#hint2
```

```{r test-check}
#solution
gradethis::grade_code()
```


### unknown (`awk`)
```{awk awk-chunk, exercise=TRUE}

```


### custom
```{R addition-custom, exercise=TRUE, exercise.cap = "My Caption Here"}

```
### custom #2
```{R addition-custom-2, exercise=TRUE, exercise.cap = "<div style='color:green;'>My Green Caption</div>"}

```

### bash
```{bash bash-chunk, exercise = TRUE}

```

### c
```{c c-chunk, exercise = TRUE}

```

### cc
```{cc cc-chunk, exercise = TRUE}

```

### coffee
```{coffee coffee-chunk, exercise = TRUE}

```

### css
```{css css-chunk, exercise = TRUE}

```

### go
```{go go-chunk, exercise = TRUE}

```

### groovy
```{groovy groovy-chunk, exercise = TRUE}

```

### haskell
```{haskell haskell-chunk, exercise = TRUE}

```

### js
```{js js-chunk, exercise = TRUE}

```

### mysql
```{mysql mysql-chunk, exercise = TRUE}

```

### node
```{node node-chunk, exercise = TRUE}

```

### octave
```{octave octave-chunk, exercise = TRUE}

```

### perl
```{perl perl-chunk, exercise = TRUE}

```

### psql
```{psql psql-chunk, exercise = TRUE}

```

### python
[...]

PR #436, opened 2020-09-29
Remove exercise icons

schloerke:
Follow-up to #429 and #431

* Removes all icons from simple icons

PR task list:
- [NA] Update NEWS
  * Going to leave this alone for now. Will remove the prior PR news entry later if necessary


R/knitr-hooks.R
```diff
@@ -317,14 +317,43 @@ install_knitr_hooks <- function() {
...
+                  "groovy" = "Groovy",
+                  "haskell" = "Haskell",
+                  "js" = "JavaScript",
+                  "mysql" = "MySQL",
+                  "node" = "Node.js",
+                  "octave" = "Octave",
+                  "psql" = "PostgreSQL",
+                  "python" = "Python",
+                  "r" = "R",
+                  "rcpp" = "R C++",
```
trestletech, line 345: Just one vote so do as you will, but... `rcpp` resonates with me in a way that `R C++` doesn't. I read the latter first and I wasn't sure what it was.

trestletech (approved):

R/knitr-hooks.R
coatless: Rcpp would be preferred to R C++. The latter doesn't exactly jive with WRE since _R_ only makes available a C API.

R/knitr-hooks.R
```diff
@@ -317,14 +317,43 @@ install_knitr_hooks <- function() {
...
+                  "rscript" = "Rscript",
```
coatless, line 346: Also, to be a bit more future proof, I think [`cpp11`](https://github.com/r-lib/cpp11) should be added to the list. 

```suggestion
                  "cpp11" = "cpp11",
                  "rscript" = "Rscript",
```

R/knitr-hooks.R
schloerke:
```suggestion
                  "rcpp" = "Rcpp",
```

schloerke:
Thank you @coatless !

PR #438, opened 2020-09-30
Colin/output styles

ColinFay:
This PR implements a new way to customize the outputs from both `{learnr}` and `{gradethis}` using exercise options.

### DIFF
--- a/R/exercise.R
+++ b/R/exercise.R
@@ -15,7 +15,6 @@ setup_exercise_handler <- function(exercise_rx, session) {
 
   # observe input
   observeEvent(exercise_rx(), {
-
     # get exercise
     exercise <- exercise_rx()
 
@@ -241,7 +240,8 @@ evaluate_exercise <- function(exercise, envir, evaluate_global_setup = FALSE) {
   # include any checker feedback with the exercise results
   exercise_result(
     feedback = checker_feedback$feedback,
-    html_output = rmd_results$html_output
+    html_output = rmd_results$html_output,
+    exercise = exercise
   )
 }
 
@@ -253,7 +253,7 @@ try_checker <- function(exercise, name, check_code, envir_result,
     get_checker_func(exercise, name, envir_prep),
     error = function(e) {
       message("Error occured while retrieving 'exercise.checker'. Error:\n", e)
-      exercise_result_error(e$message)
+      exercise_result_error(e$message, color = exercise$options$exercise.alert_color %||% "red")
     }
   )
   # If retrieving checker_func fails, return an error result
@@ -280,7 +280,7 @@ try_checker <- function(exercise, name, check_code, envir_result,
       name, paste(missing_args, collapse = "', '")
     )
     message(msg)
-    return(exercise_result_error(msg))
+    return(exercise_result_error(msg, exercise = exercise))
   }
 
   # Call the check function
@@ -288,8 +288,9 @@ try_checker <- function(exercise, name, check_code, envir_result,
     do.call(checker_func, args),
     error = function(e) {
       msg <- paste("Error occurred while evaluating", sprintf("'%s'", name))
+      cli::cat_rule("./R/exercise.R#292")
       message(msg, ": ", conditionMessage(e))
-      exercise_result_error(msg)
+      exercise_result_error(msg, exercise = exercise)
     }
   )
   # If checker code fails, return an error result
@@ -298,7 +299,7 @@ try_checker <- function(exercise, name, check_code, envir_result,
   }
   # If checker doesn't return anything, there's no exercise result to return
   if (length(feedback)) {
-    exercise_result(feedback)
+    exercise_result(feedback, exercise)
   } else {
     feedback
   }
@@ -395,7 +396,7 @@ render_exercise <- function(exercise, envir, envir_prep) {
     # make the time limit error message a bit more friendly
     pattern <- gettext("reached elapsed time limit", domain = "R")
     if (grepl(pattern, msg, fixed = TRUE)) {
-      return(exercise_result_timeout())
+      return(exercise_result_timeout(exercise = exercise))
     }
     if (length(exercise$error_check)) {
       # Run the condition through an error checker (the exercise could be to throw an error!)
@@ -409,10 +410,11 @@ render_exercise <- function(exercise, envir, envir_prep) {
         engine = exercise$engine
       )
       if (is_exercise_result(checker_feedback)) {
+        cli::cat_rule("./R/exercise.R#414")
         return(checker_feedback)
       }
     }
-    exercise_result_error(msg)
+    exercise_result_error(msg, exercise = exercise)
   })
 
   if (is_exercise_result(output_file)) {
@@ -436,8 +438,9 @@ render_exercise <- function(exercise, envir, envir_prep) {
       message = "The submitted code didn't produce a visible value, so exercise checking may not work correctly.",
       type = "warning", correct = FALSE
     )
+    browser()
     html_output <- htmltools::tagList(
-      feedback_as_html(invisible_feedback),
+      feedback_as_html(invisible_feedback, exercise),
       html_output
     )
   }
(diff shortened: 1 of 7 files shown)


ColinFay:
Examples at: 

+ http://connect.thinkr.fr/learnr-colored-output/ 
+ http://connect.thinkr.fr/learnr-colored-output-gradethis/

R/options.R
```diff
@@ -18,29 +18,59 @@
 #' @param exercise.completion Use code completion in exercise editors.
 #' @param exercise.diagnostics Show diagnostics in exercise editors.
 #' @param exercise.startover Show "Start Over" button on exercise.
+#' @param exercise.alert_class,exercise.gradethis_info_color,exercise.gradethis_success_color,exercise.gradethis_warning_color,exercise.gradethis_danger_color Slug for the CSS class for `{learnr}` and `{gradethis}` message.
+#' It can be one of `red`, `orange`, `purple`, `blue`, `violet`, `yellow`, `pink`, `green`,
+#' or `grey`. You can also implement your own CSS rule, in that case you need to define a
+#' class that starts with `alert-` (for example `alert-rainbow`).
```
schloerke, line 24: In learnr, I'd like to get away from color and instead provide the raw class name.

* `exercise.alert_class = "alert-red"` CSS class value used when an error is produced
* `exercise.info_class = "alert-info"` CSS class value used to display informational feedback
* `exercise.success_class = "alert-success"` CSS class value used to display successfully submitted code
* `exercise.warning_class = "alert-warning"` CSS class value used to display warning feedback
* `exercise.danger_class = "alert-danger"` CSS class value used to display unsuccessful  feedback

Maybe add something about how these values can be set to some predefined classes. Ex: `alert-red`, `alert-orange`, `alert-purple`, `alert-blue`, `alert-violet`, `alert-yellow`, `alert-pink`, `alert-green`, or `alert-grey`. Or you may use your own CSS class name.

schloerke (changes requested):

R/exercise.R
```diff
@@ -463,41 +468,120 @@ exercise_code_chunks <- function(exercise) {
...
+        list(
+          feedback = feedback,
+          error_message = error_message,
+          timeout_exceeded = timeout_exceeded,
+          html_output = html_output
+        ),
+        class = "learnr_exercise_result"
+      )
+    )
+  }
```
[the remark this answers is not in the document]
ColinFay: @schloerke when the code sent via Run is empty, the exercise object is NULL, so if I put a stop here it will make the tutorial crash (hence my return of an empty list): 

Screenshot 2020-10-01 at 10 44 51
Screenshot 2020-10-01 at 10 45 01

ColinFay:
@trestletech @schloerke @dcossyleon

As of our last discussion, this is now the default behavior: 

- Run Code always output the console
- If this output is an error, a colored square is shown
- Submit answer can use specific colors for the output, and the feedback and console output can be both turned off (default is display them both)

=> `{learnr}`: http://connect.thinkr.fr/learnr-colored-output/
=> `{gradethis}`: http://connect.thinkr.fr/learnr-colored-output-gradethis/

trestletech:
Looking good to me! @dcossyleon I can't remember where we netted out in the last meeting, but I notice that in the current demo learnr exercise syntax errors have their output highlighted in an alert background. I think that's helpful since we don't have the ability to print error text in right like we could in the IDE. But I remember you had reservations. Did we decide we did or did not want that?

R/exercise.R
```diff
@@ -463,41 +468,130 @@ exercise_code_chunks <- function(exercise) {
...
+  feedback = NULL,
+  html_output = NULL,
+  error_message = NULL,
+  timeout_exceeded = FALSE,
+  exercise = NULL
+) {
+
+  # When `exercise` is empty, we return a list of NULL
+  # `exercise` is NULL when there was no code sent (i.e the
+  # student hasn't entered any code)
```
schloerke, line 506:
```suggestion
  # When `exercise` is empty, we return a list of as.is values
```

R/exercise.R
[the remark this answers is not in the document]
schloerke: That is an awesome screenshot!

Can we explicitly set `exercise = NULL` near line 189 in exercise.R to be explicit on what is happening? Thank you!

```r
    return(exercise_result(html_output = " ", exercise = NULL))
```

R/exercise.R
```diff
@@ -463,41 +468,130 @@ exercise_code_chunks <- function(exercise) {
...
+
   feedback <- feedback_validated(feedback)
-  feedback_html <- feedback_as_html(feedback)
+  feedback_html <- feedback_as_html(feedback, exercise)
+
+  if (
+    is.null(exercise$check) &&
+    is.null(exercise$code_check)
+  ){
+    exercise.submitted_feedback <- TRUE
```
schloerke, line 528: When "run"ning code, should this value be set to `FALSE`? (So that there is no feedback)

```suggestion
    exercise.submitted_feedback <- FALSE
```

R/feedback.R
```diff
@@ -52,17 +55,47 @@ feedback_as_html <- function(feedback) {
...
+    role = "alert",
+    tags$pre(
+      message
+    )
+
+  )
+
+
+
 }
```
schloerke, line 101:
```suggestion
  )
}
```

schloerke (changes requested):

schloerke:
The code changes look great! Just a few minor changes left. 😄

dcossyleon:
> Looking good to me! @dcossyleon I can't remember where we netted out in the last meeting, but I notice that in the current demo learnr exercise syntax errors have their output highlighted in an alert background. I think that's helpful since we don't have the ability to print error text in right like we could in the IDE. But I remember you had reservations. Did we decide we did or did not want that?

Originally I had suggested that "Run Code" always give the standard black output, but I agree now that some sort of color indication of an error would indeed be helpful. The colored background feels too bulky for "Run Code" error output, but a nice compromise would be the suggestion of using the colored outline (see below).

This is subtle enough that the focus remains on the text in the error message, while still giving a color indication that something went awry. I think the colored background should be saved for `gradethis` cases only, where human feedback is also being shown and the student is expecting a "louder" indication of success or failure.

image

dcossyleon:
For gradethis output styles:

Output that gets wrapped in `<pre>` tags, like the example, below looks great! 

Screen Shot 2020-10-07 at 11 40 32 AM

Could we apply the same spacing (and border removal) that separates the text and output for other output types (tables, images, plots, etc.)?

screenshots

R/exercise.R
ColinFay: When clicking Run Code there is no feedback shown (there used to be a message but we decided to remove it), so as far as I can tell, it doesn't need to be switched here.

ColinFay:
@dcossyleon this is implemented in the last commit: 

Screenshot 2020-10-13 at 10 51 22
Screenshot 2020-10-13 at 10 51 37

R/exercise.R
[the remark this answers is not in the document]
ColinFay: Done in last commit ✅

R/exercise.R
```diff
@@ -463,41 +468,128 @@ exercise_code_chunks <- function(exercise) {
...
+  ){
+    exercise.submitted_feedback <- FALSE
+    exercise.submitted_output <- TRUE
+  } else {
+    exercise.submitted_feedback <- exercise$options$exercise.submitted_feedback %||% TRUE
+    exercise.submitted_output <- exercise$options$exercise.submitted_output %||% TRUE
+  }
+
+
+
```
schloerke, line 534:
```suggestion
```

R/exercise.R
```diff
@@ -463,41 +468,128 @@ exercise_code_chunks <- function(exercise) {
...
+          html_output = html_output
+        ),
+        class = "learnr_exercise_result"
+      )
+    )
+  }
+
   feedback <- feedback_validated(feedback)
-  feedback_html <- feedback_as_html(feedback)
+  feedback_html <- feedback_as_html(feedback, exercise)
```
schloerke, line 520:
```suggestion
  feedback_html <- feedback_as_html(feedback, exercise = exercise)
```

R/exercise.R
```diff
@@ -436,8 +440,9 @@ render_exercise <- function(exercise, envir, envir_prep) {
       message = "The submitted code didn't produce a visible value, so exercise checking may not work correctly.",
       type = "warning", correct = FALSE
     )
+
     html_output <- htmltools::tagList(
-      feedback_as_html(invisible_feedback),
+      feedback_as_html(invisible_feedback, exercise),
```
schloerke, line 445:
```suggestion
      feedback_as_html(invisible_feedback, exercise = exercise),
```

schloerke (approved):

PR #445, opened 2020-10-23
Colin/sanity check

ColinFay:
This is a PR implementing a sanity-check for a Rmd. (WIP for now)


R/check_exercise.R
```diff
@@ -0,0 +1,407 @@
...
+
+  # Setting something to return that will be turned to FALSE
+  # if any check fail
+  all_passed <- TRUE
+
+  # Now that we have manipulated the chunks, we can build the
+  # exercise objects
+
+  for (chunk_ in chunks_){
+    # Restoring  knitr code stock
```
schloerke, line 284: Before restoring, we need to validate that the chunk is an exercise chunk.

Something like `if (!is_exercise_chunk(chunk$options)) next`

R/check_exercise.R
```diff
@@ -0,0 +1,407 @@
...
+    all_related <- grep(chunk_, names(usefull_chunks), value = TRUE)
+
+    # If there is a setup chunk, we make sure it's the first of the list
+    if (any(grepl("setup", all_related))){
+      setup_ <- grepl("setup", all_related)
+      all_related <- c(
+        all_related[setup_],
+        all_related[!setup_]
+      )
+    }
```
schloerke, line 296: Will need to recursively find all chunks. This can be done in a followup PR.

R/check_exercise.R
```diff
@@ -0,0 +1,407 @@
...
+    # - exercise$timestamp
+    # - exercise$global_setup if any
+    # - exercise$setup if any
+    # - exercise$code_check if any
+    # - exercise$chunks that contains the setup chunks
+    # - exercise$check if any
+    # - exercise$engine
+    #
+
+    # We keep the  chunk we need: setup, and solution
```
schloerke, line 335:
```suggestion
    # We keep the  chunk we need: EX-setup, and EX-solution
```

R/check_exercise.R
schloerke: This involves how `chunks_needed` is created

R/check_exercise.R
```diff
@@ -0,0 +1,407 @@
...
+  }
+
+  # Setting something to return that will be turned to FALSE
+  # if any check fail
+  all_passed <- TRUE
+
+  # Now that we have manipulated the chunks, we can build the
+  # exercise objects
+
+  for (chunk_ in chunks_){
```
schloerke, line 283:
```suggestion
  for (chunk_label in maybe_exercise_labels) {
```

R/check_exercise.R
```diff
@@ -0,0 +1,407 @@
...
+      },
+      timestamp = {
+        as.numeric(Sys.time())
+      },
+      global_setup = {
+        paste0(setup_chunk$code, collapse = "\n")
+      },
+      setup = {
+        paste0(grab_chunk("setup")[[1]]$code, collapse = "\n")
+      },
```
schloerke, line 365: This is currently not used in `evaluate_exercise()`. Can set to NULL.

```suggestion
      setup = NULL,
```

R/check_exercise.R
```diff
@@ -0,0 +1,407 @@
...
+
+    usefull_chunks[[i]]$code <- knitr::knit_code$get(
+      usefull_chunks[[i]]$params$label
+    )
+
+  }
+
+  # We extract the default setup chunk, and remove it from the list
+  setup_chunk <- usefull_chunks[["setup"]]
+  usefull_chunks[["setup"]] <- NULL
```
schloerke, line 164: Should only need to evaluate this once. Save the result into an environment

```suggestion
  setup_chunk <- usefull_chunks[["setup"]]
  usefull_chunks[["setup"]] <- NULL
  setup_chunk_envir <- eval(parse(text = setup_chunk), envir = new.env(parent = globalenv()))
```

R/check_exercise.R
```diff
@@ -0,0 +1,407 @@
...
+      },
+      engine = {
+        usefull_chunks[chunk_][[1]]$enginee %||% "r"
+      },
+      version = "1"
+    )
+
+    res <- evaluate_exercise(
+      exercise_blt,
+      envir = new.env()
```
schloerke, line 387: Use the previously source `'setup'` environment

```suggestion
      envir = setup_chunk_envir
```

R/check_exercise.R
```diff
@@ -0,0 +1,407 @@
...
+      },
+      code_check  = {
+        # TODO
+        #grab_chunk("code-check")[[1]]
+      },
+      options = {
+        usefull_chunks[chunk_][[1]]$options
+      },
+      engine = {
+        usefull_chunks[chunk_][[1]]$enginee %||% "r"
```
schloerke, line 380:
```suggestion
        usefull_chunks[chunk_][[1]]$engine %||% "r"
```

sandbox/sandbox-exos.Rmd
```diff
@@ -0,0 +1,124 @@
...
+```{r setup, include=FALSE}
+library(learnr)
+library(gradethis)
+library(nycflights13)
+gradethis_setup()
+knitr::opts_chunk$set(echo = FALSE)
+```
+
+## Flight
+
```
schloerke, line 19:
````suggestion

```{r random_chunk}
# should not be checked
1 + 1 
```

````

**Review by schloerke — CHANGES_REQUESTED** [review]
Good work!

Need to find the true setup chunks. This can be done in a followup PR.

Would like to make sure we are not using `knitr::*$get()` as much as possible (as there is no _real_ knitr environment while creating the exercise objects).

File: R/check_exercise.R [hunk 10]
```diff
@@ -0,0 +1,407 @@
...
+      }
+
+
+      # set the eval property as appropriate
+      options$eval <- exercise_eval
+      options$echo <- FALSE
+    }
+
+    # Add the options list to the chunk
+    usefull_chunks[[i]]$options <- options
```
nischalshrestha, line 266: Note: do not overwrite useful chunk options permanently. A exercise could be a setup chunk for another exercise.

schloerke: To help with this, I think we should abstract out a `check_exercise_chunk()` method that can be `lapply`ed over the exercises. 

This _inner_ should also have the ability to allow for the caller to pass in their own function value.   If an author has a bad merge, and they are using `gradethis::grade_code()`, all it would be checking is if `x == x`, not if `x` is correct. 

Both of these functions would serve great purposes. `check_exercises()` would be a quick and dirty check. `check_exercise_chunk()` could be used within CI to validate that tutorials work as expected with many different inputs.

-----------------------

Something like...

```r
check_exercise_chunk <- function(
  file,
  ..., # should be ignored
  chunks = file_to_chunks(file),
  user_code = NULL # or expression
) {
  # ....
  if (rlang::is_expression(user_code)) {
    user_code <- solution_code
  }
  # ....
  # return feedback object
}

# would use the solutions for every exercise
check_exercises("file.Rmd") 

# would use the supplied user code to 'ex1' in 'file.Rmd'
check_exercise_chunk("file.Rmd", "ex1", user_code = rlang::expr({
  # this is user code
  40 + 2
})
```

---------------------

I would be REALLY happy to have this inside `gradethis` to validate that `gradethis` is working as expected with MANY different types of testing documents.

PR #444, opened 2020-10-17
Fix exercise solution in ex-data-filter

helix84:
arr_delay is in minutes

PR #462, opened 2021-01-04
Fix hours example in ex-data-filter.Rmd

jtransue:
Fixed two errors. The error in the first exercise is the most important change because the solution is wrong (it says to do less than 2 minutes, instead of less than 120 (2 hours)). This is my first pull request ever so please forgive mistakes.

This is meant to be an improvement in the built-in RStudio Tutorial, not to the learnr package in general. 

## Pull Request

PR task list:
- [ ] Update NEWS
- [ ] Add tests (if possible)

### DIFF
--- a/inst/tutorials/ex-data-filter/ex-data-filter.Rmd
+++ b/inst/tutorials/ex-data-filter/ex-data-filter.Rmd
@@ -105,7 +105,7 @@ filter(flights, month = 1)
 
 ### Multiple tests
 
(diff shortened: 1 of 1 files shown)


schloerke:
I like the comment added compared to just the number fix in https://github.com/rstudio/learnr/pull/444/files#diff-c50e94530959fe09537639917f64160b15e4b762da07029a0256cdfee21c5884R235 . Merging

PR #516, opened 2021-04-15
Add hybrid method for tutorial storage

elimillera:
## Hybrid Storage

We discovered when we deployed learnr tutorials to a hosted environment that certain parts of the tutorial would take a long time(30-45 seconds) to initialized. This was because we were using the 'filesystem' option when storing user data and the filesystem was a lot slower as the tutorials got larger and more users accessed it at once. This also provides a fallback if the user deletes their cookies or is blocking them.

Our solution was to combine the client and filesystem methods. This has worked very well for us and we're hoping to contribute it back to the package. 

Functions in `hybrid_storage` are a combination of client and filesystem storage with the exception of `get_objects`. `get_objects` includes logic to restore the state in the filesystem to the client storage if there are zero or one items in the client storage.

PR task list:
- [x] Update NEWS
- [x] Add tests (if possible) - Tests were added however because the option requires a call to `session$request` we were unable to include them. Its possible they could be used in a future version of shiny where testServer is updated to handle this.


mstackhouse:
Closed as we have an adjustment we realized we need to implement for when users are working across multiple browsers.

PR #532, opened 2021-06-02
Fix exercise prompt in ex-data-filter.Rmd

dfailing:
Suggest changing "Submit Answer" to "Run Code" to avoid confusion in the `filter()` section.


gadenbuie (approved):
Thanks @dfailing!

PR #539, opened 2021-06-12
Make `data/` directory automatically accessible to exercises

rossellhayes:
This adds two interfaces for making files available to `learnr` tutorials.

1. Any files in a `data/` directory in the same directory as the tutorial Rmd document will be accessible to the tutorial.
2. Calling the function `use_remote_files()` in the `setup` chunk will cache files into a temporary directory and make them available to the tutorial as if they were in the `data/` directory.


gadenbuie (changes requested):
This is a great start! I really like how the interface is coming together and I think it will benefit many tutorial authors.

We'll need to think through and account for all of the various ways the underlying options can be set and that files can be added to an exercise. To help with that, I think the next step would be to add several small test tutorials in a folder at `tests/testthat/exercise-files`. Each one can demonstrate an expected use case. Actually writing the tests is going to be a little tricky, but we can work through that step together. In the mean time it will be helpful to have a few examples to test out interactively.

R/file_helpers.R
```diff
@@ -0,0 +1,122 @@
...
+#' @param ... Remote files to include in the tutorial.
+#'   Each file will be accessible at `data/<name of input>`.
+#'   Unnamed arguments have their name set by default with [basename()].
+#'
+#' @return Sets [`knitr` chunk options][knitr::opts_chunk]
+#'
+#' @export
+
+use_remote_files <- function(...) {
+  data_files <- rlang::flatten_chr(list(...))
```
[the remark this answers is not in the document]
rossellhayes: I guess I was using "remote" here to mean "not in the normal directory". This function works with both URLs and file paths outside of the data directory. Should we only allow URLs as remote files, or do you think this function could use a name change?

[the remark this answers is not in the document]
gadenbuie: I'm not sure. I can think of a few ways that local files not in the tutorial working directory can cause problems, e.g. a random file elsewhere on my machine that will be inaccessible to a deployed tutorial. But there are also "local" files that we might want to work, such as files included in other R packages and accessed with `system.file()`.

I think that once we have a few reprexes it'll be more clear what we should do with this function.

R/file_helpers.R
```diff
@@ -0,0 +1,122 @@
...
+      dir(temp_dir, recursive = TRUE, full.names = TRUE)
+    )
+    names(remote_files) <- dir(temp_dir, recursive = TRUE)
+  }
+
+  # Combine paths to local files and remote files
+  files <- c(local_files, remote_files)
+
+  # Specify full paths in chunk options
+  knitr::opts_chunk$set(exercise.files = files)
```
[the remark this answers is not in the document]
rossellhayes: I've set it up now so that if a user has a `data/` directory and uses `use_remote_files()` and manually sets the chunk option, all those files will be accessible to the user. However, I've set a warning if the chunk options are manually set, as this will lead to suboptimal performance (e.g. remote files won't be cached). The warning shows up in the R Markdown pane when the exercise is rendered with the Run button and the console when the the exercise is rendered with `run()`.

R/file_helpers.R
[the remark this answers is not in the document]
rossellhayes: For now I've set a check to ensure remote files are either URLs or package files. We can reevaluate if we find any other beneficial uses.

tests/testthat/test-file_helpers.R
```diff
@@ -0,0 +1,56 @@
...
+  expect_equal(
+    get_option_remote_files(),
+    system.file("examples", "knitr-minimal.Rnw", package = "knitr")
+  )
+
+  use_remote_files("https://covidtracking.com/api/v1/states/daily.csv")
+  expect_equal(
+    get_option_remote_files(),
+    "https://covidtracking.com/api/v1/states/daily.csv"
+  )
```
gadenbuie, line 21: Does passing this test also imply that the second call to `use_remote_files()` replaces the files previously declared by the first `use_remote_files()`?

We might need a `with_knitr_opts_chunk()` or a `with_knitr_opts_knit()` equivalent of `withr::with_options()`.

R/file_helpers.R
```diff
@@ -1,122 +1,188 @@
...
 }
 
+is_system_file <- function(path) {
+  apply(
+    as.matrix(
+      vapply(.libPaths(), grepl, logical(length(path)), path, fixed = TRUE)
+    ),
+    1,
+    any
+  )
```
gadenbuie, line 183: This is fun! But also not very readable. I played with it a bit and decided I like this version slightly better:

```suggestion
  vapply(
    lapply(path, Vectorize(grepl, "pattern"), pattern = .libPaths(), fixed = TRUE),
    any,
    logical(1)
  )
```

but...

gadenbuie: ...but I think it might be even better as a slightly recursive function

```r
is_system_file <- function(path) {
  if (length(path) > 1) {
    vapply(path, is_system_file, logical(1))
  }
  any(vapply(.libPaths(), grepl, x = path, fixed = TRUE, logical(1)))
}	
```

rossellhayes: Love this!

tests/testthat/test-exercise.R
```diff
@@ -472,4 +472,99 @@ test_that("exercise versions upgrade correctly", {
   # (this version of learnr makes a strong assumption that "label" is part of exercise)
   ex_99_broken$label <- NULL
   expect_error(upgrade_exercise(ex_99_broken))
-})
\ No newline at end of file
+})
+
+test_that("data files", {
```
gadenbuie, line 477: Could you please split up this test into individual `test_that()` calls? This would let you give each testing block a more descriptive name, e.g. `"data files in data/ are copied into exercise temp dir"`, etc.

R/use_data_dir.R
```diff
@@ -0,0 +1,44 @@
...
+    rlang::warn(paste0('The data directory "', dir, '" could not be found.'))
+  }
+
+  options(learnr.data_dir = dir)
+
+  invisible(dir)
+}
+
+copy_data_dir <- function(exercise_dir) {
+  # First check environment variable, then options(), then default to "data/"
```
gadenbuie, line 27: Let's flip this and check options -> env var -> default to `data/`

R/use_data_dir.R
```diff
@@ -0,0 +1,44 @@
...
+  source_dir <- Sys.getenv(
+    "LEARNR_DATA_DIR", unset = getOption("learnr.data_dir", default = "data")
```
gadenbuie, line 29: Let's also change this env var name to `TUTORIAL_DATA_DIR` and `tutorial.data.dir`, since this will likely be set at the _tutorial_ level.

R/use_data_dir.R
```diff
@@ -0,0 +1,44 @@
...
+  )
+
+  if (!dir.exists(source_dir)) {return(invisible(NULL))}
```
gadenbuie, line 32: We should report this failure, probably with a warning

R/use_data_dir.R
```diff
@@ -0,0 +1,44 @@
...
+
+  dest_dir <- file.path(exercise_dir, "data")
+  dir.create(dest_dir)
+
+  if (!dir.exists(dest_dir)) {
+    rlang::abort('The user-facing data directory was not created successfully.')
```
gadenbuie, line 38: Where is this error raised? Will it stop a running tutorial?

gadenbuie, line 38: Turns out this error is presented to the end-user in the tutorial. I think this wording might make it more clear to users that it isn't their fault and give a better hint to authors.

```suggestion
    rlang::abort("An error occurred: we weren't able to create the data directory for this exercise.")
```

R/use_data_dir.R
gadenbuie: I can see why it's currently invisible -- if there isn't a `data/` directory we don't want to add a lot of noise. So we should put some thought into how we can surface warnings when its clear that the user would want to see them without creating  noise for tutorials without data folders.

R/initialize.R
```diff
@@ -40,6 +40,13 @@ initialize_tutorial <- function() {
       singleton = TRUE
     )
 
+    # # Make files accessible to tutorial
+    # rmarkdown::shiny_prerendered_chunk(
+    #   'server-start',
+    #   'learnr:::prepare_data_files()',
+    #   singleton = TRUE
+    # )
+
```
gadenbuie, line 49:
```suggestion

```

R/use_data_dir.R
```diff
@@ -0,0 +1,44 @@
...
+  if (!dir.exists(dir)) {
+    rlang::warn(paste0('The data directory "', dir, '" could not be found.'))
+  }
+
+  options(learnr.data_dir = dir)
+
+  invisible(dir)
+}
+
+copy_data_dir <- function(exercise_dir) {
```
gadenbuie, line 26: What do you think of making `source_dir` an argument of `copy_data_dir()` and adding a `data_dir` argument to `evaluate_exercise()`?

gadenbuie (changes requested):

R/use_data_dir.R
rossellhayes: My thinking here was that the envvar would be set when the tutorial is deployed on a remote server, and in that case the option set by the author wouldn't necessarily be a valid path anymore. But that makes me wonder if using `options()` is the wrong approach for this. How would the remote deployment know which directories it needs? Maybe non-default directories should be handled in the YAML rather than the setup chunk?

R/use_data_dir.R
rossellhayes: I think this makes sense, especially with my earlier comment about potentially using the YAML

R/use_data_dir.R
rossellhayes: I've changed this to return invisibly if "data/" doesn't exist, but to raise an error if a different directory is specified (by envvar or `options()`) and that directory doesn't exist (5b4a0ee).

R/use_data_dir.R
gadenbuie: I think that options > env var is the right order. The R process can read the env var, so the author who sets the option can also read and use the env var. In other words, letting the env var overwrite the option removes flexibility. I have the same wonder about whether or not we should use `options()`, though.

R/use_data_dir.R
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  file.copy(dir(source_dir, full.names = TRUE), dest_dir, recursive = TRUE)
+
+  return(invisible(dest_dir))
+}
+
+err_msg <- function(...) {
+  paste("An error occurred:", ...)
+}
```
gadenbuie, line 55: This is one of those times when it's okay to copy/paste the same thing in a couple places; I'm not sure there's much if any benefit of having this in a function.

R/use_data_dir.R
```diff
@@ -0,0 +1,55 @@
...
+
+copy_data_dir <- function(exercise_dir) {
+  # First check environment variable, then options(), then default to "data/"
+  source_dir <- Sys.getenv(
+    "TUTORIAL_DATA_DIR", unset = getOption("tutorial.data.dir", default = "data")
+  )
+
+  if (!dir.exists(source_dir)) {
+    if (source_dir == "data") return(invisible(NULL))
```
gadenbuie, line 33: I think a better signal might be to set the fallback default value of `source_dir` to `""`. Then if `identical(source_dir, "")` check if `data/` exists in the current working directory, and if it doesn't you can exit quietly. Otherwise, exit loudly if we can't find `data/` or proceed as normal. The benefit is that if a user explicitly sets `option(tutorial.data.dir = "data")`, we'll give an error.

rossellhayes: I like this solution, implemented in 6672a55

R/exercise.R
```diff
@@ -300,6 +302,16 @@ evaluate_exercise <- function(exercise, envir, evaluate_global_setup = FALSE) {
   dir.create(exercise_dir)
   on.exit(unlink(exercise_dir), add = TRUE)
 
+  # Copy files from data directory into exercise
+  if (is.null(data_dir)) {
+    # First check environment variable, then default to "data/" if it exists
+    source_dir <- Sys.getenv(
```
gadenbuie, line 308: I think we lost the `tutorial.data_dir` option in this step. We definitely want to look at R option -> env var -> final default

R/exercise.R
```diff
@@ -300,6 +302,16 @@ evaluate_exercise <- function(exercise, envir, evaluate_global_setup = FALSE) {
...
+      "TUTORIAL_DATA_DIR",
+      unset = if (dir.exists("data")) {"data"} else {""}
```
gadenbuie, line 310:
```suggestion
      unset = if (dir.exists("data")) "data" else ""
```

R/data_dir.R
```diff
@@ -0,0 +1,28 @@
+copy_data_dir <- function(source_dir, exercise_dir) {
+  if (identical(source_dir, "")) {
+    return(invisible(NULL))
+  }
+
+  if (!dir.exists(source_dir)) {
+    rlang::abort(
+      paste("An error occured:",
+            "we weren't able to find the data directory for this exercise."),
```
gadenbuie, line 9:
```suggestion
      "An error occurred with the tutorial: the data directory does not exist.",
```

R/data_dir.R
```diff
@@ -0,0 +1,28 @@
...
+      class = "learnr.missing_source_dir"
```
gadenbuie, line 10:
```suggestion
      class = "learnr_missing_source_data_dir"
```

R/data_dir.R
```diff
@@ -0,0 +1,28 @@
...
+    )
+  }
+
+  dest_dir <- file.path(exercise_dir, "data")
+  dir.create(dest_dir)
+
+  if (!dir.exists(dest_dir)) {
+    rlang::abort(
+      paste("An error occurred:",
+            "we weren't able to create the data directory for this exercise."),
```
gadenbuie, line 20:
```suggestion
      "An error occurred with the tutorial: we weren't able to create the data directory for this exercise.",
```

R/data_dir.R
```diff
@@ -0,0 +1,28 @@
...
+      class = "learnr.missing_dest_dir"
```
gadenbuie, line 21:
```suggestion
      class = "learnr_missing_dest_data_dir"
```

R/exercise.R
rossellhayes: I may have misunderstood the goal of the latest change. If we're requiring authors to keep files in the `"data/"` directory, when would this option be set?

R/data_dir.R
```diff
@@ -0,0 +1,28 @@
...
+    )
+  }
+
+  file.copy(dir(source_dir, full.names = TRUE), dest_dir, recursive = TRUE)
```
gadenbuie, line 23: This works, but wow I can't believe that there isn't a base R way to copy a directory!

R/exercise.R
gadenbuie: Somewhere along the way, most likely by an external evaluator. The env var is too high-level in that case.

gadenbuie: Let's call the option `tutorial.data_dir`

rossellhayes: That makes sense!

R/data_dir.R
rossellhayes: I know, it's there in `fs`!

R/exercise.R
```diff
@@ -304,10 +304,13 @@ evaluate_exercise <- function(
...
-    source_dir <- Sys.getenv(
-      "TUTORIAL_DATA_DIR",
-      unset = if (dir.exists("data")) "data" else ""
+    # First check options(), then environment variable, then default to "data/"
+    source_dir <- getOption(
+      "tutorial.data_dir",
+      default = Sys.getenv(
+        "TUTORIAL_DATA_DIR",
+        unset = if (dir.exists("data")) {"data"} else {""}
+      )
```
gadenbuie, line 313: The braces snuck back in and it's okay not to name the second argument of `getOption()` and `Sys.getenv()`

```suggestion
    source_dir <- getOption(
      "tutorial.data_dir",
      Sys.getenv("TUTORIAL_DATA_DIR", if (dir.exists("data")) "data" else "")
```

tests/testthat/test-exercise.R
```diff
@@ -557,7 +557,81 @@ test_that("error if env var directory does not exist", {
...
     expect_error(
       evaluate_exercise(ex, new.env(), evaluate_global_setup = TRUE),
-      class = "learnr.missing_source_dir"
+      class = "learnr_missing_source_data_dir"
     )
   })
 })
+
+test_that("alternate data directory specified with `options()`", {
+  withr::with_tempdir({
```
gadenbuie, line 566: Thanks to the pattern you started in #544 https://github.com/rstudio/learnr/blob/bcd5dc3ba831bf44d255dc2643201610161dda32/R/exercise.R#L315-L318

I realized we can avoid putting the whole test in `with_tempdir()`

```suggestion
  withr::local_dir(withr::local_tempdir())
```

tests/testthat/test-exercise.R
```diff
@@ -557,7 +557,81 @@ test_that("error if env var directory does not exist", {
...
+        writeLines("MODIFIED", "data/test.txt")
+        readLines("data/test.txt")
+      ',
+      global_setup = 'options(tutorial.data_dir = "nested/structure")'
+    )
+    output <- evaluate_exercise(ex, new.env(), evaluate_global_setup = TRUE)
+    expect_match(output$html_output, "MODIFIED", fixed = TRUE)
+    expect_match(readLines("data/test.txt"), "DEFAULT", fixed = TRUE)
+    expect_match(readLines("nested/structure/test.txt"), "NESTED", fixed = TRUE)
+  })
```
gadenbuie, line 600: To finish the thought above...

```suggestion
  })
```

tests/testthat/test-exercise.R
gadenbuie: This is totally worth the effort -- it makes it so much easier to step through the test interactively. withr is really cool 🤯

tests/testthat/test-exercise.R
```diff
@@ -557,7 +557,81 @@ test_that("error if env var directory does not exist", {
...
+    dir.create("data")
+    writeLines("DEFAULT", "data/test.txt")
+    dir.create("nested/structure/data", recursive = TRUE)
+    writeLines("NESTED", "nested/structure/test.txt")
+
+    ex <- mock_exercise(
+      user_code = 'readLines("data/test.txt")'
+    )
+    output <- evaluate_exercise(ex, envir = new.env())
+    expect_match(output$html_output, "DEFAULT", fixed = TRUE)
```
gadenbuie, line 576: A little trick that we didn't talk about (because I forgot about it until I read through a few tests) is that you can set `check = TRUE` in `mock_exercise()` this uses a debugging exercise checker that returns a value in the `$feedback` item of the exercise result. 

In that flow, you can access any of the objects that were made available to the exercise checker function via `$feedback$checker_args`. So, for example, you can directly access the result of running the user code like this:

```suggestion
    ex <- mock_exercise(
      user_code = 'readLines("data/test.txt")',
      check = TRUE
    )
    res <- evaluate_exercise(ex, envir = new.env())
    expect_equal(res$feedback$checker_args$last_value, "DEFAULT")
```

You can use this trick to run code at various stages of the exercise evaluation that you can then access via the `envir_prep` or `envir_result` environments that are passed to the checker function.

tests/testthat/test-exercise.R
```diff
@@ -557,7 +557,81 @@ test_that("error if env var directory does not exist", {
...
+    expect_match(readLines("data/test.txt"), "DEFAULT", fixed = TRUE)
+    expect_match(readLines("nested/structure/test.txt"), "NESTED", fixed = TRUE)
```
gadenbuie, line 578: What do these tests tell us? (I don't think we gain any information from them and we can probably remove them and the others like them.)

tests/testthat/test-exercise.R
```diff
@@ -557,7 +557,81 @@ test_that("error if env var directory does not exist", {
...
+    ex <- mock_exercise(
+      user_code = '
+        writeLines("MODIFIED", "data/test.txt")
+        readLines("data/test.txt")
+      ',
+      global_setup = 'options(tutorial.data_dir = "nested/structure")'
+    )
+    output <- evaluate_exercise(ex, new.env(), evaluate_global_setup = TRUE)
+    expect_match(output$html_output, "MODIFIED", fixed = TRUE)
+    expect_match(readLines("data/test.txt"), "DEFAULT", fixed = TRUE)
```
gadenbuie, line 598: This test makes sense: we're expecting that `data/test.txt` here refers to our original `data/` directory. (Make a mental bookmark, we'll need to revisit this test when we work out directory precedence.)

tests/testthat/test-exercise.R
rossellhayes: Elegant!

gadenbuie:
@rossellhayes I had a few small changes to propose and in the process I merged main into this branch. It turns out that our options-protection work misses options that are `NULL`. In other words `withr::local_options()` doesn't reset options that were `NULL`, so our tests are now failing until we fix that.

```r
options(thing = NULL)
withr::with_options({
  options(thing = "bad")
})
getOption("thing")
#> "bad"
```

To be fair, this same issue will happen with the standard base R pattern too:

```
options(thing = NULL)
local({
  opts <- options()
  on.exit(options(opts), add = TRUE)
  options(thing = "bad")
})
getOption("thing")
#> "bad"
```

We'll have to write a custom `options_restore()` function that _does_ handle `NULL` options.

rossellhayes:
It looks like this is a known issue: https://github.com/r-lib/withr/issues/57
Will Landau proposes a solution here: https://github.com/r-lib/withr/issues/57#issuecomment-332007780

gadenbuie (approved):
This looks great and I'm really excited for this feature to land in learnr. I'm approving the PR but we have a little bit more to do in terms of documentation.

- [ ] Add an entry in NEWS under **New Features**
- [ ] Document the new behavior in the learnr docs webpage. I think it makes sense to add a new section in `docs/exercises.Rmd` after [Exercise Setup](https://rstudio.github.io/learnr/exercises.html#Exercise_Setup).
- [ ] Once the above are done and the PR is merged, we'll need to follow up in any related issues to let everyone know about the new feature.

NEWS.md
```diff
@@ -20,6 +20,7 @@ learnr (development version)
 * Previously, when a question submission was reset, it would be recorded as a `"question_submission"` event with the value `reset=TRUE`. Now it a separate event, `"reset_question_submission"`. ([#398](https://github.com/rstudio/learnr/pull/398))
 * Added a new `polyglot` tutorial to learnr. This tutorial displays mixing R, python, and sql exercises. See [`run_tutorial("polyglot", "learnr")`](https://learnr-examples.shinyapps.io/polyglot) for a an example. ([#397](https://github.com/rstudio/learnr/pull/397))
 * Text throughout the learnr interface can be customized or localized using the new `language` argument of `tutorial()`. Translations for English and French are provided and contributes will be welcomed. Read more about these features in `vignette("multilang", package = "learnr")`. ([#456](https://github.com/rstudio/learnr/pull/456), [#479](https://github.com/rstudio/learnr/pull/479))
+* Tutorials can attach a file directory for use within exercises. Files in a "data/" directory within the same directory as the tutorial R Markdown document will be available from within exercises. ([#539](https://github.com/rstudio/learnr/pull/539))
```
gadenbuie, line 23:
```suggestion
* When a "data/" directory is found in the same directory as the tutorial R Markdown document, it is now automatically made available within exercises. An alternative directory can be specified using the `tutorial.data_dir` global option. ([#539](https://github.com/rstudio/learnr/pull/539))
```

R/exercise.R
```diff
@@ -789,6 +792,27 @@ merge_options <- function(preserved_opts, inherited_opts, static_opts = list())
   opts[!grepl("^exercise", names(opts))]
 }
 
+local_restore_options <- function(.local_envir = parent.frame()) {
+  options <- options()
+  withr::defer(restore_options(options), envir = .local_envir)
```
schloerke, line 797: While the original code works, seeing a variable `options` quickly has me think that the value is the `options()` function. Let's change it to something other than `options`.

```suggestion
  opts <- options()
  withr::defer(restore_options(opts), envir = .local_envir)
```

R/exercise.R
```diff
@@ -789,6 +792,27 @@ merge_options <- function(preserved_opts, inherited_opts, static_opts = list())
...
+
+  envvars <- Sys.getenv()
+  withr::defer(restore_envvars(envvars), envir = .local_envir)
+}
```
schloerke, line 801: Let's turn this into two methods to not overload the name / functionality

```suggestion
}

local_restore_envvars <- function() {
  envvars <- Sys.getenv()
  withr::defer(restore_envvars(envvars), envir = .local_envir)
}
```

R/exercise.R
```diff
@@ -434,8 +438,7 @@ get_checker_func <- function(exercise, name, envir) {
 
 render_exercise <- function(exercise, envir) {
   # Protect global options and environment vars from modification by student
-  withr::local_options(list())
-  withr::local_envvar(as.list(Sys.getenv()))
+  local_restore_options()
```
schloerke, line 441:
```suggestion
  local_restore_options()
  local_restore_envvars()
```

R/exercise.R
```diff
@@ -283,10 +283,11 @@ required_names <- c("code", "label", "options", "chunks", require_items)
...
 #   evaluators, if they choose to use this function, might want to include the
 #   global setup.
-evaluate_exercise <- function(exercise, envir, evaluate_global_setup = FALSE) {
+evaluate_exercise <- function(
+  exercise, envir, evaluate_global_setup = FALSE, data_dir = NULL
+) {
   # Protect global options and environment vars from permanent modification
-  withr::local_options(list())
-  withr::local_envvar(as.list(Sys.getenv()))
+  local_restore_options()
```
schloerke, line 290:
```suggestion
  local_restore_options()
  local_restore_envvars()
```

R/exercise.R
schloerke: If you'd like to make a wrapper around these two functions, I'm ok with it.

schloerke (approved):
LGTM given cosmetic suggestions!

PR #612, opened 2021-11-22
Add random phrases in German

NinaCorrelAid:
Adds random praise/encouragement in German


gadenbuie (approved):
Thanks again @NinaCorrelAid!

PR #761, opened 2023-02-16
Set tab size in Ace editor to 2

rossellhayes:
The Ace editor used in learnr tutorials defaults to using 4 spaces for tabs, but Tidyverse style recommends two spaces.

@gadenbuie I think I made the needed change in `tutorial.js`, but I'm not sure how `build.js` works, could you help me out with that part?


gadenbuie (approved):
Looks good, thanks! To update the learnr javascript resources you can run the following in the Terminal in the learnr repo:

```bash
# install the required js libraries (call once)
npm install

# re-build learnr's js sources
npm run build
```

PR #772, opened 2023-03-21
Add empty last line to ex-data-mutate example

davidkane9:
I just wanted to fix lack of an empty last line since it causes an annoying warning when I parsermd::parse_rmd() the file. I did not think is necessary to add tests or update the docs.

PR task list:
- [x ] Update NEWS
- [ ] Add tests (if possible)


gadenbuie (approved):
Thanks @davidkane9! And you're right, we don't need a NEWS item for this change, so I reverted that commit.

PR #776, opened 2023-04-03
Move `curl` to "Suggests"

rossellhayes:
The `curl` package is only required to enable external evaluators. Since external evaluators aren't needed for most `learnr` users, this PR moves `curl` from "Imports" to "Suggests".


gadenbuie (approved):
LGTM thanks!

PR #777, opened 2023-04-04
Make 'continue' button customisable in sub-topics

dave-mills:
## Customise 'continue' button text

This PR adds the ability to specify custom 'continue' button text for each subtopic with a `data-continue-text` attribute. It uses the same syntax as for adding the [progressive reveal](https://pkgs.rstudio.com/learnr/v0.10.1.9021/articles/exercises.html?q=progress#progressive-reveal) option to individual topics: 

```r
### Subtopic Heading {data-continue-text: "Show Next Section"}
```

Any subtopics without this attribute will still use the default "Continue" text, as defined by the translation files. 

### Rationale
We use learnr a lot for writing interactive modules for online or mixed-learning R, statistics and data management courses. Often, we use progressively revealed subtopics for self-assessed quizzes, and we wanted a way to replace the 'Continue' text with e.g. "Show Solution", or "Next Question". I made a hack-y version for us with a custom JavaScript file, but I thought it'd be nicer to have a built-in option.

### Concerns
I understand that this may be a very niche feature, so you may not want to include it for that reason. I also realise that the custom text is not translatable with this setup, but given that each document only uses 1 language at a time right now, I figured that making the custom text translatable wasn't vital. But if/when learnr is updated to allow multiple languages at the same time this would need fixing. 

PR task list:
- [x] Update NEWS
- [ ] Add tests (if possible)

(btw - I didn't know if/how this could be testable. I'm not very familiar with R package development, so happy to add tests if needed or tidy up other package-dev things!)

Thanks for your time, and thanks for such a great package for building training resources!


dave-mills:
Example of an Rmd file with this feature: 

````r
---
title: "Example custom button text"
output:
  learnr::tutorial:
    progressive: yes
runtime: shiny_prerendered
description: Example learnr tutorial with custom continue buttons
---

```{r setup, include=FALSE}
library(learnr)
```

## Topic 1

### Question 1 {data-continue-text="Show Solution"}

Question text goes here.

### Solution 1 {data-continue-text="Next Question"}

Solution text goes here.

### Question 2 {data-continue-text="Show Solution"}

Question 2 text.

### Solution 2

The next button has the 'default' text for your langauge.

## Topic 2

etc...
````

gadenbuie (approved):
Thanks! I refactored the code slightly and added a manual test. Overall this looks great! It'd be nice to have this documented somewhere in the learnr website, but I can add that later before the next CRAN release.

PR #788, opened 2023-05-23
First version translated into Spanish from the firt 5 tutorials in package learnr (referring to the v1 of r4ds book)

xavidp:
First version translated into Spanish from the firt 5 tutorials in package learnr (referring to the v1 of r4ds book)
I couldn't finish the check package task since for some unknown reason, no pdflatex is available (apparently) even after installing tinytex, or removing tinytex and installing miktex-portable and restarting R each time. 

PR task list:
- [X] Update NEWS
- [ ] Add tests (if possible)
- [X] Update documentation with `devtools::document()`


xavidp:
I added also some new translation to the learnr UI into Spanish

